PR #15, opened 2024-07-03
IMU code

### LINKED ISSUE #9: Implement new IMU Code
There is a new IMU on the Internal FC Board V1a. We need to add the libraries and initializations.

TheSilentDefender:
Added code that reads gyro, acceleration, and temperature data. `c.accel` pulls data from the accelerometer while `c.imu_accel` pulls acceleration data from the IMU. Resolves #9.


CircuitPy/FC_Board/lib/pysquared.py
```diff
@@ -145,10 +145,10 @@ def __init__(self):
 
         # Initialize Accelerometer
```
Mikefly123, line 146: The intended use of the lsm303 is actually for its onboard magnetometer not the accelerometer. This should be changed around to instead return mag values rather than redundant access values.

CircuitPy/FC_Board/lib/pysquared.py
```diff
@@ -272,6 +272,34 @@ def reset_vbus(self):
...
+
+    @property
+    def imu_accel(self):
+        try:
+            return self.imu.acceleration
+        except Exception as e:
+            self.debug_print('[ERROR][ACCEL]' + ''.join(traceback.format_exception(e)))
+
+    @property
+    def temp(self):
```
Mikefly123, line 290: If you're planning to call the onboard temp sensor you should call it something like ```imu_temp``` to differentiate it from other sensors that might be on the board.

CircuitPy/FC_Board/lib/adafruit_lsm303_accel.mpy
Mikefly123, line 1: This needs to be the mag library rather than the accel library!

Mikefly123 (changes requested):
Hey @TheSilentDefender, the code looks good! The main thing is we actually want to use the LSM303 for mag values rather than accel. Take a look at my comments and ping me again once you've made some changes!

PR #17, opened 2024-07-11
Fixed PySquared Import Issues

Mikefly123:
This should solve most of the issues arising from inconsistent pin names in the new firmware!


Mikefly123:
Wait there are missing things here still

PR #36, opened 2024-10-10
Attempt to re-implement fsk operation using adafruit_rfm modules

Camerooooon:
This pull request is a work in progress.

### DIFF
--- a/FC_Board/lib/functions.py
+++ b/FC_Board/lib/functions.py
@@ -29,7 +29,7 @@ class functions:
             "Hey Its pretty cold up here, did someone forget to pay the electric bill?"
         ]
         self.last_battery_temp = 20
-        self.callsign = "Callsign"
+        self.callsign = "KN6YZZ"
         self.state_bool = False
         self.face_data_baton = False
         self.detumble_enable_z = True
@@ -59,7 +59,9 @@ class functions:
         message = f"{self.callsign} " + str(msg) + f" {self.callsign}"
         self.field.Beacon(message)
         if self.cubesat.f_fsk:
-            self.cubesat.radio1.cw(message)
+            pass
+            # TODO: Reimplement method
+            #self.cubesat.radio1.cw(message)
         if self.cubesat.is_licensed:
             self.debug_print(f"Sent Packet: " + message)
         else:
@@ -95,7 +97,9 @@ class functions:
         self.field = Field.Field(self.cubesat, self.debug)
         self.field.Beacon(lora_beacon)
         if self.cubesat.f_fsk:
-            self.cubesat.radio1.cw(lora_beacon)
+            pass
+            # TODO: Reimplement method
+            #self.cubesat.radio1.cw(lora_beacon)
         del self.field
         del Field
 
@@ -117,7 +121,7 @@ class functions:
                 f"UT:{self.cubesat.uptime}",
                 f"BN:{self.cubesat.c_boot}",
                 f"MT:{self.cubesat.micro.cpu.temperature}",
-                f"RT:{self.cubesat.radio1.former_temperature}",
+                f"RT:{self.cubesat.radio1.temperature}",
                 f"AT:{self.cubesat.internal_temperature}",
                 f"BT:{self.last_battery_temp}",
                 f"AB:{int(self.cubesat.burned)}",
@@ -137,12 +141,13 @@ class functions:
                 + str(self.state_list)
                 + f"{self.callsign}"
             )
-            if self.cubesat.f_fsk:
-                self.cubesat.radio1.cw(
-                    f"{self.callsign} Yearling^2 State of Health 1/2"
-                    + str(self.state_list)
-                    + f"{self.callsign}"
-                )
+            # TODO: Reimplement method
+            #if self.cubesat.f_fsk:
+            #    self.cubesat.radio1.cw(
+            #        f"{self.callsign} Yearling^2 State of Health 1/2"
+            #        + str(self.state_list)
+            #        + f"{self.callsign}"
+            #    )
             self.state_bool = True
         else:
             self.field.Beacon(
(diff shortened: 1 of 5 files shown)


Camerooooon:
Currently I am unable hear any tx from the satellite using a radio despite the output of the "send" function (See Field.py line 46) being True. I'm not sure why.

Also @Mikefly123 would you say that this would be a good way of going about adding in the adafruit_rfm libraries? I noticed the old (now removed in this PR) pysquared_rfm9x.py is a modified version of the adafruit library but I don't think we need any modifications so I simply copied the library from the adafruit library.

Mikefly123:
Cruising through the changes that you've made it seems like a reasonable way of trying to get this working. Ideally we would want to implement some kind of system for switching the radio config values depending on if we want LoRa or FSK mode rather than just nuking all of the LoRa config stuff. 

@Camerooooon do you have a SDR that you can hookup to monitor the frequency you're trying to transmit on and watch for activity on the waterfall to see if anything is coming out of the radio at all?

Camerooooon:
> Cruising through the changes that you've made it seems like a reasonable way of trying to get this working. Ideally we would want to implement some kind of system for switching the radio config values depending on if we want LoRa or FSK mode rather than just nuking all of the LoRa config stuff.
> 
> @Camerooooon do you have a SDR that you can hookup to monitor the frequency you're trying to transmit on and watch for activity on the waterfall to see if anything is coming out of the radio at all?

I believe the adafruit_rfm [rfm9xfsk](https://github.com/proveskit/circuitpy_flight_software/pull/36/files#diff-b477fa10ab323b9501b25cfca4b256e12823e95e8812198509c7f0e6ea1bf138) module that I am implementing also has support for LoRa via the rfm9x module which is a drop in replacement for the rfm9xfsk one which is why I am commenting out the LoRa config for now. I will need to figure the best way to implement switching between radio modes in the future.

As for the SDR, I was using a simple quad band FM transceiver to monitor the frequency and it was not making any noise. From my experience with fsk it should sound like a squawking noise which I didn't hear. I will try an SDR tomorrow since I ran out of time today.

FC_Board/lib/functions.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, cubesat):
             "Hey Its pretty cold up here, did someone forget to pay the electric bill?"
         ]
         self.last_battery_temp = 20
-        self.callsign = "Callsign"
+        self.callsign = "KN6YZZ"
```
Camerooooon, line 32: Need to remember not to commit this before merge.

FC_Board/lib/pysquared.py
```diff
@@ -75,7 +75,7 @@ def __init__(self):
         self.debug = True  # Define verbose output here. True or False
         self.legacy = False  # Define if the board is used with legacy or not
         self.heating = False  # Currently not used
-        self.is_licensed = False
+        self.is_licensed = True
```
Camerooooon, line 78: Need to remember not to commit this before merge.

Mikefly123:
Hey @Camerooooon! Just checking if you managed to check it out using an SDR.

Camerooooon:
> Hey @Camerooooon! Just checking if you managed to check it out using an SDR. 

I havent had time yet, will try on Monday.

Camerooooon:
Hello @Mikefly123 I think work is finally back on track as far as this pr.

Using an SDR to try and view the signal I noticed a very large momentary blip of signal spanning across the entire spectrogram when the signal ends the transmission. In addition there is a very faint signal around 4 mhz above the desired transmission frequency that lasts throughout the transmission. I'm not sure why there is such odd behavior. Luckily @DavidUniack just told me he will lend me some test boards so I can get more work done.

Camerooooon:
@Mikefly123 
20241110_16h24m27s_grim_anotated

Camerooooon:
After looking more into I think that this might just be random noise as a result of the two devices being connected to the same computer and I think that the "Line of noise spanning all frequencies" might be the device attempting to transmit and failing.

Mikefly123:
Hm, pretty interesting @Camerooooon can you share two things to give us some more context? 
- What does your SDR look like when you transmit LoRa packets? 
- What is the length of the test message you're trying to send? (short strings like "Hello World" might be hard to see compared to a 250 Byte String)

Camerooooon:
> Hm, pretty interesting @Camerooooon can you share two things to give us some more context?
> 
>     * What does your SDR look like when you transmit LoRa packets?
> 
>     * What is the length of the test message you're trying to send? (short strings like "Hello World" might be hard to see compared to a 250 Byte String)

The LORA looks nothing like what I was seeing in the image above, it was a strong signal which was being picked up across multiple frequencies. As for the length of the message I am sending the full status report message so I don't think the length is the issue. After more testing I noticed that the default baud rate was very high and after lowering it in the most recent commit my SDR now shows a good clean signal when transmitting FSK! I tried decoding the FSK with the SDR with no success but I imagine if I use another satellite as a client I could successfully decode the FSK signal.

Camerooooon:
Here is a screenshot of the FSK tranmission
20241112_13h20m12s_grim

Mikefly123:
Oh great @Camerooooon! Have you tried to decode the FSK using a second flight controller board? The `adafruit_rfm` library should natively support receiving and directly decoding with the hardware. I've personally found that using SDRs to decode digital modulations to be incredibly frustrating.

Camerooooon:
> Oh great @Camerooooon! Have you tried to decode the FSK using a second flight controller board? The `adafruit_rfm` library should natively support receiving and directly decoding with the hardware. I've personally found that using SDRs to decode digital modulations to be incredibly frustrating. 

Yeah that's what I'm going to try when I get a chance.

Camerooooon:
This should be ready for testing. Allows for switching between LoRa and FSK mode based on the f_fsk value.

Alternating between FSK and LoRa every few days is not implemented on this PR.

@Mikefly123 this is ready for test & merge.

Mikefly123 (approved):
LGTM!

PR #42, opened 2024-10-19
Convert self.hardware to Ordered Dict

### LINKED ISSUE #37: Change c.hardware to be an Ordered Dict
# What is the Proposed Change? 
In CircuitPython 9 the Adafruit Team added the ability to have an ordered dictionary data structure through the `collections` library. 

https://docs.circuitpython.org/en/latest/docs/library/collections.html

Currently the `self.hardware` in the `pysquared.py` is an unordered dict that will print the status of the cubesat hardware in a semi-random order at the end of the `pysquared` initializations. By making it an ordered dict it will just generally look a little nicer! 

# Reference Code: 
```py         
self.hardware = {
            "I2C0": False,
            "SPI0": False,
            "I2C1": False,
            "UART": False,
            "IMU": False,
            "Mag": False,
            "Radio1": False,
            "SDcard": False,
            "NEOPIX": False,
            "WDT": False,
            "TCA": False,
            "CAN": False,
            "Face0": False,
            "Face1": False,
            "Face2": False,
            "Face3": False,
            "Face4": False,
        }
```

dbgen1:
Resolves #37.

All references to the hardware field have been checked and should work properly. I also slightly re-ordered the dictionary to align with the order of initialization.


Mikefly123:
Thanks for submitting this! I will test it on our Engineering Model tomorrow and if it runs well I will approve the pull request!

Mikefly123:
Ran it and LGTM! We might want to check down the line for any edge case issues with unpacking the dict into state of health stuff but it shouldn't be an issue.

Mikefly123 (approved):

PR #47, opened 2024-11-01
38 add microcontroller clock functions

Mikefly123:
Added some new calls! In pysquared you can now `c.turbo` to change the clock speed from 62.5Mhz (when it is `False`) to 125Mhz (when it is `True`). Also made it so that we boot into 62.5Mhz clock for power saving. 

Also added a fix for TCA `try_lock` hangs when the camera inits.


Lex-ari:
lgtm

Lex-ari (commented):
lgtm lgtm

PR #26, opened 2024-07-29
Split detumble into a package, add tests and type hints

nateinaction:
Work in progress. Need to test on hardware.

This PR:
- Refators detumble into separate module
- Adds type hints
- Adds first unit tests
- Adds automated pipeline to run tests

PR #57, opened 2024-12-24
Created unit tests for detumble.py

TaylorGilg:
- Created unit tests for functions dot_product(), x_product(), and magnetorquer_dipole() in detumble.py (detumble.py is identical for Batt_Board and FC_Board) with realistic argument inputs 
Bigger context:
- do_detumble() (in functions.py in FC_Boards and battery_functions.py in Batt_Board) calls the magnetorquer_dipole() method (that calls the other methods dot_product and x_product) with tuple arguments of magnetic field strength at x,y,& z axis and angular velocities at x,y,& z axis. The method returns a list value for dipole_out (to be used to inform actuators on how much force must be applied for detumbling)


nateinaction:
Holy moly the first unit tests in the repo! I've done some work on adding automated testing to the repo in #26 and chose pytest because instead of the built-in unittest library because it seemed more widely used and was faster because it could run tests in parallel ([here's a review comparing them](https://blog.jetbrains.com/pycharm/2024/03/pytest-vs-unittest/)). 

Could you switch these tests over to use pytest and then reuse some of the automation introduced in PR #26 to have it run automatically on commit? Pulling over those changes might be as simple as `git checkout origin/python-type-declarations -- .github/workflows/ci.yaml .vscode/settings.json Makefile README.md requirements.txt`

If you change the requirements file don't forget to rerun your pip install!

nateinaction:
This is looking good! Since there are two "separate" code bases in this repo `FC_Board` and `Batt_Board` it might be helpful to have separate test suites for both until we combine everything into a single board. As it turns out, putting the test files inside of `/FC_Board` helps with the import issue we're seeing in CI too! I put up a few changes against your branch here: https://github.com/proveskit/circuitpy_flight_software/pull/62

If you like them, feel free to include them!

nateinaction (approved):

PR #58, opened 2024-12-24
Implemented Type Hinting in the functions.py file and detumble.py

blakejameson:
I defined type hints for the function signatures in the functions.py file and detumble.py. The type hints should allow for a better understanding of each function. 

Something to note is that while I did implement type hinting for the functions, I also provided type hints for the member variables in the constructor function of the 'functions' class. That can be seen starting in line 32. I wasn't suggested to do this but thought this could provide clarity later on to the developer whenever they reference these fields.


FC_Board/lib/functions.py
```diff
@@ -15,25 +15,32 @@
...
+    pass
+from pysquared import Satellite
+
 
 class functions:
 
-    def debug_print(self, statement):
+    def debug_print(self, statement: str) -> None:
         if self.debug:
             print(co("[Functions]" + str(statement), "green", "bold"))
```
nateinaction, line 30: It looks like `statement` has a type conversion to `str` on line 30. It almost looks like a bug but since type hints aren't enforced at runtime this feels fair for now.

nateinaction (commented):
I have checked a good number of these type hints and they appear to be correct. Sadly type hinting is not enforced at runtime so this should be safe to implement at anytime. All the same, can you verify that you've attempted to run this on the FC board and didn't run into any new errors?

FC_Board/lib/functions.py
blakejameson: I got held up on this function a while because of the conversion there on line 30. I saw that every time the function is called, there is a string passed. However, line 30 made me question if I should change statement to just be of type 'Any'. This was a tough one

blakejameson:
> I have checked a good number of these type hints and they appear to be correct. Sadly type hinting is not enforced at runtime so this should be safe to implement at anytime. All the same, can you verify that you've attempted to run this on the FC board and didn't run into any new errors?

 @nateinaction 
 Yes, I can confirm that I ran this on the FC board and didn't experience errors

nateinaction (approved):

PR #60, opened 2024-12-24
Resolving 3 intellisense issues

nateinaction:
Intellisense changes: added comments where resolution was not immediately available, renamed imports where it fixed issues, removed unused imports where possible. Payload file is also deleted because it is not used and contains an import that will fail at runtime since it does not exist.

This resolves only 3 squigglies but I also [learned about mpy files](https://docs.micropython.org/en/latest/reference/mpyfiles.html) and am starting to think through how to best have library .py files for development but possibly use mycropython's `mpy-cross` tool for more easly loading code onto the board.

- [x] Tested on board


FC_Board/lib/Big_Data.py
```diff
@@ -1,6 +1,4 @@
 from debugcolor import co
-import time
-import traceback
```
nateinaction, line 3: Imports unused

FC_Board/lib/pysquared.py
```diff
@@ -9,20 +9,26 @@
...
+    mount,
+    umount,
+    VfsFat,
+)  # circuitpython module will need stub/mock for IDE intellesense, Issue #27
+import digitalio, sdcardio  # circuitpython module will need stub/mock for IDE intellesense, Issue #27
+from os import stat, mkdir, chdir
+from bitflags import bitFlag, multiBitFlag
+from micropython import (
+    const,
+)  # circuitpython module will need stub/mock for IDE intellesense, Issue #27
```
nateinaction, line 24: several unused imports removed, black linter caused multiline split

TaylorGilg (approved):

PR #64, opened 2024-12-28
Implemented config files

Alandlt15:
commandsConfig.py: holds dictionary of commands

config.json: holds other configuration data (eg. cubesat name, radio callsign, jokes, etc)


Mikefly123:
Looks like a good start to me! I think that these init variables from `pysquared.py` should be next up to be added to the config files: 

```py
        """
        Big init routine as the whole board is brought up. Starting with config variables.
        """
        self.debug = True  # Define verbose output here. True or False
        self.legacy = False  # Define if the board is used with legacy or not
        self.heating = False  # Currently not used
        self.orpheus = True  # Define if the board is used with Orpheus or not
        self.is_licensed = True

        """
        Define the normal power modes
        """
        self.NORMAL_TEMP = 20
        self.NORMAL_BATT_TEMP = 1  # Set to 0 BEFORE FLIGHT!!!!!
        self.NORMAL_MICRO_TEMP = 20
        self.NORMAL_CHARGE_CURRENT = 0.5
        self.NORMAL_BATTERY_VOLTAGE = 6.9  # 6.9
        self.CRITICAL_BATTERY_VOLTAGE = 6.6  # 6.6
        self.vlowbatt = 6.0
        self.battery_voltage = 3.3  # default value for testing REPLACE WITH REAL VALUE
        self.current_draw = 255  # default value for testing REPLACE WITH REAL VALUE
        self.REBOOT_TIME = 3600  # 1 hour
        self.turbo_clock = False
```

Alandlt15:
Ended up finishing the toml file for configuration. But when I connect to the board and the countdown starts, I get a "module named 'tomllib' not found" error. 

I did some digging and found that tomllib, which is a library I need to parse data from the config file, was introduced in Python 3.11, which I know the actual board is about 6 versions behind. 

My question is, would I be able to use a third party library called 'toml' I would simply need to run the command pip install toml I think, and parse exactly how I was attempting with the 'tomllib'. But I'm not too sure if that's a valid solution to this. 

Otherwise, I believe my only option is to keep using json.

Mikefly123:
Hey @Alandlt15 I think this one of those things that is a mismatch between mainstream Python and CircuitPython.

It would appear from the [CircuitPython docs](https://learn.adafruit.com/circuitpython-octoprint-controller-and-monitor/create-your-settings-toml-file) that they only natively support accessing a `settings.toml` using the `os` module. This is the example that they give: 

```py
import os

print(os.getenv("test_variable"))
```
I'm not sure if this allows us to actually write values to the `settings.toml` during runtime or if it treats these variables as static. You may want to do some tests with your own board! If the implementation ends up being too clunky compared to using a `.json` then we can just go back to doing that.

Alandlt15:
I ended up settling with JSON. One reason behind this is because CircuitPython "only supports a subset of the full TOML specification" leading to only integers and strings being supported. (I have floats, booleans, and lists inside my config file). 

I did give the third party TOML a try but at first I kept getting "module named 'toml' not found". I kinda figured that toml was not installed directly on the board, so I decided to manually add the toml folders to the lib directory on the FC_Board directory. This seemed to solve that, however I then encountered another "module named 'datetime' not found" issue. 

Additionally, there are bugs that accept invalid TOML files. It is mentioned those will be fixed with no timeframe specified. This might not seem like a big deal to files that are valid TOML, but I believe it is safer and better to settle with JSON for configuration. 

I've yet to test if values can be changed at runtime, I think it is achievable, but still need to do some research and testing on that. 

Nevertheless, the current JSON config implementation works on the board!

Mikefly123:
Cool! Thanks for the testing and write up. Do you think we should keep this PR open until we finish migrating everything, we want to be in the `config.json` or should we do a merge in the interim and start new PR(s) for additional features? 

Like you mentioned in your last comment, being able to overwrite the values in the `json` during runtime will be really important for over the air updates. Also, you may want to look into storing a copy of the `config.json` on the SD card as a backup.

Alandlt15:
Now that I'm thinking about it, it would be wise to just leave it open until the migration is done. This would allow me to definitely implement value changes at runtime, and discover if other variables should be added to the config. 

Thank you for taking the time to review and comment!

nateinaction (approved):
This looks like a good start. I think you should consider merging and continuing to improve/modify in small PRs based on the mainline rather than continuing to build on this PR and merging a larger PR later. PRs of this size are easier to review.

My suggested next steps are:
1. Only read the config file once on application startup: This will be a challenging refactor but will help us implement the ability to update config values in-flight because the control flow will be more easily understandable.
2. Create a config class which, on initialization, reads the config file and places all the values into private member variables. Provide a public getter method on the class so other parts of the program can access the config values. This structure will allow you to more easily control how config values are changed in the future by providing setter method where you can implement how configs values are changed. If we want to follow the FPrime model and allow values to be set temporarily before being made permanent on restart, this is probably the right structure.
3. Create a public setter method that allows changing the member variables but don't write that variable to the config file.
4. Create a public writer method that allows modifying the config file with the updated config value.

Alandlt15:
That sounds like a plan to me, I fixed some conflicts, black linter was giving me some trouble but I fixed it. Did one last test with the board and all is working fine. 

I will merge shortly, thank you all for your time and advice!

PR #66, opened 2024-12-28
DRAFT POC Reorganize and use unmodified upstream libs

nateinaction:
This PR attempts to:
- Removed unused code
- Organize the repo so absolute imports are usable everywhere
- Provide a mechanism to download specific library versions
- Remove modified upstream libraries
- Provide easy to use developer experience with help dialog in the Makefile

This PR resolves all IntelliSense issues except ones requiring circuitpython stubs. This PR has been tested on the board and appears to work. This reorg brings down the amount of code we need to maintain from ~24k lines to ~3k lines.


nateinaction:
Broken up and implemented in #70 #73 #76 #67 and #81. Closing.

PR #70, opened 2025-01-04
Replace any unmodified libraries with their upstream versions

nateinaction:
This PR replaces all unmodified python libraries with their upstream counterparts. I have reviewed changes for each replaced library and believe there were no breaking changes.

3 libraries have been excluded from this PR because they contain modifications:
- `adafruit-circuitpython-lsm6ds`
- `adafruit-circuitpython-mcp2515`
- `adafruit-circuitpython-rfm`

I have been able to fix each of these but will submit those in a separate PR.

Before merge:
- [x] Test on board: The output on the FC board is identical to `main`


blakejameson (approved):
Ran the code on board and output was identical to that of the main branch. 
Incredible job with all the work done here Nate

PR #73, opened 2025-01-05
Finish replacing libs with their upstream version

nateinaction:
**Summary:**
Follow up to #70. This PR removes the remaining libraries from the FC_Board directory. These libraries contained modifications which have now been included in the pysquared code. In this PR, you can go commit by commit to see the exact changes needed to get each library removal working.

**Testing:**
I have verified that the board is showing the same output as on the main branch.


blakejameson (approved):
Tested on my board and had same output as in main

PR #76, opened 2025-01-06
DRAFT Reorg 1: Move FC_Board contents to repo root

nateinaction:
DRAFT: Not ready for review


Mikefly123:
Hey Nate! Since we are going to a single microcontroller solution for the flight software, I do think it makes sense to eliminate the federated structure of having separate folders for the code that goes on each board. Something I wanted to add in just for us to think about some more is the developer experience of actually getting the files onto the board whenever we want to update. 

I think the most comfortable thing for non-software-oriented people to do in this regard is a drag and drop of the files onto the CircuitPy external drive. Whereas I have usually seen folks with a software background prefer to use the command line to copy the files they want over. Maybe it could make sense to setup some kind of script that helps with that process? Or do you see installing new code on the boards to mainly be through a release system where they unzip a folder with all the relevant code and drop that in instead?

nateinaction:
@Mikefly123 I've been thinking through this too. I think you're right that we'll want separate workflows for developers vs. satellite operators. I suspect we'll have some kind of `make install` or `make sync` command in the Makefile for devs which could potentially automatically identify the board and install the code. For everyone else we'll probably want to cut releases and make the files available from the Github interface. Let's keep talking about it and iterating on the best user experience!

blakejameson (approved):
Tested on the v4c board and saw same output as main. Great job Nate!

Issue #79, opened 2025-01-07
Remove can_bus From the Codebase 


Mikefly123:
# Why is This Change Being Made?
A CAN bus transceiver was introduced in the V1.5 PROVES Kit as a means of facilitating board to board communication between the RP2040 mounted on the battery board and the RP2040 mounted on the flight controller board. This never really panned out, with the preferred method of doing board to board communication being UART for lower power consumption and simplicity. As a result in PROVES Kit V2 we will be phasing out use of the `can_bus`. 

Due to recurring problems with the `can_bus` (as seen in #71) it may be more convenient at this point to toss the `can_bus` out of the code base rather than backtracking to fix it just to remove it later. 

# What Needs to Be Done? 
The `can_bus` needs to be purged from the code base. This includes mentions of it in `pysquared.py`, `functions.py` and libraries and helper files such as `adafruit_mcp2515` and `can_bus_heplper.py`. Attached as some (but not all) examples of where this code lives in the codebase: 

In `pysquared.py`:
```py
        """
        CAN Transceiver Initialization
        """
        try:
            self.spi0cs2: digitalio.DigitalInOut = digitalio.DigitalInOut(
                board.SPI0_CS2
            )
            self.spi0cs2.switch_to_output()
            self.can_bus: CAN = CAN(self.spi0, self.spi0cs2, loopback=True, silent=True)
            self.hardware["CAN"] = True
            # TODO(nateinaction): Accessing private method. Ability to sleep the canbus should be made in upstream. Search for this comment to find other usages.
            self.can_bus._set_mode(0x20)

        except Exception as e:
            self.debug_print(
                "[ERROR][CAN TRANSCEIVER]" + "".join(traceback.format_exception(e))
            )
```
In `functions.py`: 
```py
    def safe_sleep(self, duration: int = 15) -> None:
        self.debug_print("Setting Safe Sleep Mode")

        # TODO(nateinaction): Accessing private method. Ability to sleep the canbus should be made in upstream. Search for this comment to find other usages.
        self.cubesat.can_bus._set_mode(0x20)
```

blakejameson:
@Mikefly123 
<img width="318" alt="can_bus_helper" src="https://github.com/user-attachments/assets/d99a9562-e12f-4e24-97bb-a7ceeb11572e" />

[...]

PR #80, opened 2025-01-07
Removed CAN bus, can_bus_helper, adafruit_mcp2515 FC_Board code from repo

blakejameson:
I removed the usage of the adafruit_mcp2515 library for the FC Board. 

As seen in the issue linked below, adafruit_mcp2515 is used predominantly in the Batt_Board code. As the Batt_Board code will soon be removed from the repo, with @Mikefly123 's approval, I didn't deal with CAN aka adafruit_mcp2515 for this section.

This PR is in regard to the following issue:
https://github.com/proveskit/circuitpy_flight_software/issues/79

I ran this code on the v4c and v4b boards and experienced similar output as the main branch code


Alandlt15 (approved):
Great work, files look good, just tested on my board and it works like a charm!

nateinaction (approved):
Tested on board. Looks good!

PR #81, opened 2025-01-07
Use absolute imports everywhere, all pysquared files into single lib dir

nateinaction:
This PR moves to using absolute imports to resolve intellisense issues in the IDE. It also reorganizes the pysquared library files into a single directory. I'm hoping this makes it really easy for folks to know what files we care about and gives a sense of how the library should interact with the rest of the codebase.

Testing:
I have tested this change on the board and confirm the output looks like `main`


blakejameson (approved):
Tested on v4c board and worked the same as main. Excellent job

PR #84, opened 2025-01-08
Removing side effects from pysquared.py

cosmiccodon:
Removes the side effects of importing import pyquared.py. Tested on v4c board.

### DIFF
--- a/lib/pysquared/pysquared.py
+++ b/lib/pysquared/pysquared.py
@@ -131,16 +131,14 @@ class Satellite:
         """
         Setting up data buffers
         """
-        # TODO(blakejameson): is this line in use? Remove if not?
+        # Data_cache, filenumbers, image_packets, and send_buff are variables that are not used in the codebase. They were put here for Orpheus last minute.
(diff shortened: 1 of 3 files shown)


lib/pysquared/pysquared.py
```diff
@@ -131,16 +131,14 @@ def __init__(self) -> None:
         """
         Setting up data buffers
         """
-        # TODO(blakejameson): is this line in use? Remove if not?
+        # Data_cache, filenumbers, image_packets, and send_buff are variables that are not used in the codebase. They were put here for Orpheus last minute.
+        # We are unsure if these will be used in the future, so we are keeping them here for now. EvanOrtiz BlackJameson 01/07/2025
```
nateinaction, line 135:
```suggestion
        # TODO(cosmiccodon/blakejameson):
        # Data_cache, filenumbers, image_packets, and send_buff are variables that are not used in the codebase. They were put here for Orpheus last minute.
        # We are unsure if these will be used in the future, so we are keeping them here for now.
```

nateinaction (approved):

PR #85, opened 2025-01-08
added 3 lines back

Alandlt15:
Added 3 lines of code back as well as the "Camera Initializations" comment. Originally, those lines where part of that init section (can remove if requested). Although, this comment distinguishes the code from previous and proceeding initializations.

### DIFF
--- a/lib/pysquared/pysquared.py
+++ b/lib/pysquared/pysquared.py
@@ -435,6 +435,13 @@ class Satellite:
         self.scan_tca_channels()
 
         """
+        Camera Initializations
+        """
+        if self.f_fsk:
+            self.debug_print("Next restart will be in LoRa mode.")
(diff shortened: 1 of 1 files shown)


TaylorGilg (dismissed):
I ran the code on my board and got our usual same output. 👍

lib/pysquared/pysquared.py
```diff
@@ -434,6 +434,13 @@ def __init__(self) -> None:
         """
         self.scan_tca_channels()
 
+        """
+        Camera Initializations
+        """
```
nateinaction, line 439:
```suggestion
```
I'm not sure why this was in the "Camera Initializations" section but this code modifies the radio so let's at least remove the comments.

nateinaction (approved):

PR #86, opened 2025-01-09
Bitflagrollover

cosmiccodon:
The bit rollover function is redundant, as it automatically resets to 0 after reaching 255. I searched through the entire directory for bit flag rollover implementations and deleted them.


nateinaction:
Opting for #87

PR #87, opened 2025-01-09
multiBitFlag rollover to 0 once max is reached

cosmiccodon:
Currently the `multiBitFlag` class is capable of counting from 0 to the maximum value of the number of bits requested. Once at the maximum value it would not change if incremented.

This PR changes the behavior so that the `multiBitFlag` class always rolls over to 0 after reaching its maximum. We have checked the variables assigned as the `multiBitFlag` class and believe this is not a breaking change. It's different behavior that we need to expect when reading satellite logs but no functionality should break.

### DIFF
--- a/lib/pysquared/pysquared.py
+++ b/lib/pysquared/pysquared.py
@@ -82,7 +82,7 @@ class Satellite:
             print(co("[pysquared]" + str(statement), "green", "bold"))
 
     def error_print(self, statement: Any) -> None:
-        self.c_error_count: multiBitFlag = +1  # Limited to 255 errors
+        self.c_error_count: multiBitFlag += 1  # Limited to 255 errors
         if self.debug:
             print(co("[pysquared]" + str(statement), "red", "bold"))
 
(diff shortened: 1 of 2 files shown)


lib/pysquared/pysquared.py
```diff
@@ -82,9 +82,7 @@ def debug_print(self, statement: Any) -> None:
             print(co("[pysquared]" + str(statement), "green", "bold"))
 
     def error_print(self, statement: Any) -> None:
-        self.c_error_count: multiBitFlag = (
-            self.c_error_count + 1
-        ) & 0xFF  # Limited to 255 errors
+        self.c_error_count: multiBitFlag = +1  # Limited to 255 errors
```
blakejameson, line 85: To be transparent, I need to read more through the bitflags.py to understand its purpose and functionality,

However, I do see that with this line, self.c_error_count will be set to 1 everytime and never increment. Perhaps you meant to do +=1 ?

Mikefly123: Yeah I think that's a typo. In Python you want to use +=1 to increment a variable.

Alandlt15:
```suggestion
        self.c_error_count: multiBitFlag += 1  # Limited to 255 errors
```

nateinaction (approved):
Good teamwork here. Going to merge this in so I can rebase the bigflag refactor PR.

PR #89, opened 2025-01-10
Update Makefile and Readme with new development experience

nateinaction:
## Summary

This PR updates the `Makefile` with new targets to help developers.

Running `make` takes the place of all of our setup instructions. It will setup the venv, download all dependencies, install the pre-commit hook.

You can also deploy to the board just by running `make install`

`make help` displays more of what you can do!

```sh
make help

Usage:
  make <target>
  help             Display this help.

Development
  download-libraries  Download the required libraries
  fmt              Lint and format files
  test             Run tests
  install          Install the project onto a connected PROVES Kit use `BOARD_MOUNT_POINT` to specify the mount point
  clean           Remove all gitignored files such as downloaded libraries and artifacts

Build
  build            Build the project, store the result in the artifacts directory
```

## Testing
No testing required, this is a local/pipeline only change.

FYI @Mikefly123 we were talking about something like this for devs in a previous PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: all
-all: download-libraries help
+all: venv download-libraries pre-commit-install help
 
 .PHONY: help
 help: ## Display this help.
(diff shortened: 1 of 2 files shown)


blakejameson (approved):
Ran all the commands on my device and had success. Great job Nate

PR #90, opened 2025-01-10
Replace black with ruff, now linting imports so removed unused from requirements.txt

nateinaction:
## Summary
Only human made changes are in the requirements.txt and the pre-commit-hook.yaml. Everything else is automated.

I replaced black linter with [ruff](https://docs.astral.sh/ruff/) which is faster and is capable of linting imports 🙏 which is something I have and love in my main language, golang.

## Testing
Tested on main board of behavioral changes


Alandlt15 (approved):
I can confirm this works. I created a local branch off main, and merged into into add-ruff-2 branch. Got this output:

image

PR #93, opened 2025-01-11
Non-volatile memory counter and flag refactor with unit tests

nateinaction:
## Summary
This PR replaces the `bitflags` module with an `nvm` module that is designed to help with interactions with non-volatile memory (nvm) on the microcontroller. This change started with making a hardware abstraction layer for NVM interactions so we could write unit tests for the `bitflags` module. Most of the `bitflags.multibitflag` package was tested and working in [e576ae8](https://github.com/proveskit/circuitpy_flight_software/pull/93/commits/e576ae85c62e7148186a66682dc0f289f6f1f7ce)

After getting that far, we worked to reduce the features implemented by the `multiBitFlag` class to only what the current version of the flight software uses. We ended up with a simple `Counter` class with two methods `get()` and `increment()`.

Also we built a stub for `ByteArray` which is the interface definition used to back CircuitPython's NVM implementation. I have attempted to contribute this upstream: https://github.com/adafruit/Adafruit_CircuitPython_Typing/pull/46

Next @TaylorGilg converted the `bitFlag` class to a class named `Flag` which offers two methods `get()` and `toggle()` and added tests.

## Testing
Added unit tests
Tested on board


lib/pysquared/pysquared.py
```diff
@@ -539,22 +520,6 @@ def turbo(self, value: bool) -> None:
...
-        self.f_burnarm: bitFlag = value
-
-    @property
-    def burned(self) -> bitFlag:
-        return self.f_burned
-
-    @burned.setter
-    def burned(self, value: bitFlag) -> None:
-        self.f_burned: bitFlag = value
-
```
nateinaction, line 557: Note to reviewers, these fns were unused.

TaylorGilg: Note: All but nvm registers _BOOTCNT, _ERRORCNT, and _FLAG were removed due to not being able to find any further usage of them across the repo.

TaylorGilg (approved):
Looked over all the changes and everything looks good.

PR #94, opened 2025-01-16
Scaled Config

Alandlt15:
Built a Config class to initialize variables found in pysquared.py, functions.py, and cdh.py from the json file. Instantiation of this class starts at main.py. 

The class parses the json file and places values inside a private member dictionary '_config'. It has getter and setter functions. The setter functions set values temporarily by only replacing the value inside of '_config' , and the 'saver' functions (commented out) are going to solve the issue of permanent values by writing the modified value back into the json file. 

This structure follows the FPrime Model. More info about that here under 'Parameter Database': https://nasa.github.io/fprime/UsersGuide/user/cmd-evt-chn-prm.html

I also converted the cdh.py file into a class so as to be able to receive the Config object from functions.py. I also added some type hinting inside cdh.py.


Mikefly123:
Interesting work @Alandlt15! On the skim of your requested changes they looks pretty good to me. 

Would you say that the main objective of this update is to consolidate access to the `config.json` to minimize the risk of conflicts if multiple classes are trying to read / write in the same runtime?

lib/pysquared/pysquared.py
```diff
@@ -822,3 +814,7 @@ def new_file(self, substring: str, binary: bool = False) -> Union[str, None]:
                 return None
         else:
             self.debug_print("[WARNING] SD Card not initialized")
+
+
+# print("Initializing CubeSat")
+# cubesat = Satellite()
```
nateinaction, line 820:
```suggestion
```
Lines inadvertently added back.

lib/pysquared/Config.py
```diff
@@ -0,0 +1,80 @@
...
+        return self._config[floatVariable]
+
+    def getBoolValue(self, boolVariable: str) -> bool:
+        return self._config[boolVariable]
+
+    def getListValue(self, listVariable: str) -> list[str]:
+        return self._config[listVariable]
+
+    def getCommands(self) -> dict:
+        return commandsConfig.commands
```
nateinaction, line 47: I don't think we will expect the commands to be configurable so it feels out of place here. Instead it looks like it might be better placed as a member variable of the new cdh class. I would remove the `commandsConfig` file and stick it in the class before the init:
```py
class CommandDataHandler:
    commands = ...
```

What do you think?

lib/pysquared/Config.py
```diff
@@ -0,0 +1,80 @@
...
+        return self._config[intVariable]
+
+    def getFloatValue(self, floatVariable: str) -> float:
+        return self._config[floatVariable]
+
+    def getBoolValue(self, boolVariable: str) -> bool:
+        return self._config[boolVariable]
+
+    def getListValue(self, listVariable: str) -> list[str]:
+        return self._config[listVariable]
```
nateinaction, line 44: I like this for type hinting purposes but since there is no type enforcement at runtime I'm a little dubious of the value add over a single `def get(self, key: str) -> Union(str, int, float, bool, list...)`. The single function would make this module easier to maintain in the long run. Not a blocker but something to think on and keep an eye on after you get this version merged.

In the near term, what do you think of this naming modification?
```suggestion
    def getStr(self, key: str) -> str:
        """Gets a string value from the config dictionary"""
        return self._config[key]

    def getInt(self, key: str) -> int:
        """Gets an int value from the config dictionary"""
         return self._config[key]

    def getFloat(self, key: str) -> float:
        """Gets a float value from the config dictionary"""
        return self._config[key]

    def getBool(self, key: str) -> bool:
        """Gets a bool value from the config dictionary"""
        return self._config[key]

    def getList(self, key: str) -> list[str]:
        """Gets a list value from the config dictionary"""
        return self._config[key]
```

lib/pysquared/Config.py
```diff
@@ -0,0 +1,80 @@
...
+        self._config[key] = strValue
+
+    def setIntValue(self, key: str, intValue: int) -> None:
+        self._config[key] = intValue
+
+    def setFloatValue(self, key: str, floatValue: float) -> None:
+        self._config[key] = floatValue
+
+    def setBoolValue(self, key: str, boolValue: bool) -> None:
+        self._config[key] = boolValue
```
nateinaction, line 63: Mirroring my comment on the get functions:
```suggestion
    def setStr(self, key: str, value: str) -> None:
        """Sets the string value in the config dictionary
        Does not save value to disk, will not persist through reboots
        """
        self._config[key] = value

    def setInt(self, key: str, value: int) -> None:
        """Sets the string value in the config dictionary
        Does not save value to disk, will not persist through reboots
        """
        self._config[key] = value

    def setFloat(self, key: str, value: float) -> None:
        """Sets the string value in the config dictionary
        Does not save value to disk, will not persist through reboots
        """
        self._config[key] = value

    def setBool(self, key: str, value: bool) -> None:
        """Sets the string value in the config dictionary
        Does not save value to disk, will not persist through reboots
        """
        self._config[key] = value
```

lib/pysquared/pysquared.py
```diff
@@ -4,11 +4,10 @@
 CircuitPython Version: 9.0.0
 Library Repo:
 
-* Author(s): Nicole Maggard, Michael Pham, and Rachel Sarmiento
+* Author(s): Nicole Maggard, Michael Pham, and Rachel Sarmientoc
```
nateinaction, line 7: Was this name change intentional?

nateinaction (commented):
This feels really close. Thank you for taking the time to refactor and inject the config down into the objects that need it rather than reading from the json in multiple locations!

I've added a comment about the current form of the commandsConfig abstraction as well as some minor comments on naming. Please let me know what you think!

lib/pysquared/pysquared.py
Alandlt15: Not intentional.

lib/pysquared/Config.py
Alandlt15: I agree that the commands are more than likely not going to be up for configuration. Your suggested change seems great to me since this will bring back the isolation to the commands. This also ensures they are only used exclusively within the cdh.py file.

lib/pysquared/Config.py
Alandlt15: Yup this sounds good to me. One of the things I was aiming for here was keeping the code as clean and simple as possible. Your suggested function does a good job at this.

Alandlt15:
@Mikefly123 absolutely, the Config object takes care of that by getting injected in main, and passed down to where it is needed.  And more than likely, the 'write' utility will be implemented inside of the files that use the config object to initialize values. But, this implementation definitely needs more planning and careful consideration.

lib/pysquared/functions.py
```diff
@@ -32,7 +32,7 @@ def debug_print(self, statement: Any) -> None:
         if self.debug:
             print(co("[Functions]" + str(statement), "green", "bold"))
 
-    def __init__(self, cubesat: Satellite) -> None:
+    def __init__(self, cubesat: Satellite, Config: Config) -> None:
```
nateinaction, line 35:
```suggestion
    def __init__(self, cubesat: Satellite, config: Config) -> None:
```

Looking for consistent capitalization for variable names.

requirements.txt
```diff
@@ -1,3 +1,2 @@
-adafruit-circuitpython-typing==1.11.2
```
nateinaction, line 1: Looks like inadvertent deletion from when we were playing around in the lab.

lib/pysquared/Config.py
```diff
@@ -0,0 +1,92 @@
...
+
+Also it allow values to be set temporarily using setter
+function, and values can be set permanently using the
+saver functions. Following the FPrime model.
+"""
+
+import json
+
+
+class Config:
```
nateinaction, line 14: Yes, class names should be capitalized. Keep this but use lowercase for the filename.

nateinaction (commented):
Minor comments on consistent capitalization and an accidental line deletion.

lib/pysquared/cdh.py
```diff
@@ -1,168 +1,172 @@
...
-                    print("arg decoding error: {}".format(e))
-            if cmd in commands:
+        self._repeat_code: str = Config.getStr("repeat_code").encode("utf-8")
+        print(f"Super secret code is: {self._super_secret_code}")
+
+    ############### hot start helper ###############
+    def hotstart_handler(self, cubesat, msg) -> None:
+        # try
+        try:
+            cubesat.radio1.node = cubesat.cfg["id"]  # this sat's radiohead ID
```
Mikefly123, line 33: Hey @Alandlt15 this is unrelated to the changes you made, but I noticed that on Line 33 and 34 here the `hotstart_handler()` is calling a `cubesat.cfg` which no longer exists in the current implementation of `pysquared.py`. 

We could make a separate pull request to fix this, but perhaps we can just absorb the fix in this PR since you're already updating `cdh.py`?

lib/pysquared/Config.py
Alandlt15: Got it! Making these changes as we speak.

lib/pysquared/cdh.py
Alandlt15: Yup I can make the changes on this PR!

nateinaction (approved):

PR #95, opened 2025-01-17
Bit Flag & Boot Count Refactor & Pytesting

TaylorGilg:
Removed unused bitflag.py code (nateinaction) and simplified functoriality:
- get() -> bool: Check if specific bit in a specific byte (held in byte array of non-volatile memory) is on (1 -> True) or off (0 -> False)
- toggle(value: bool) -> None: Set specific bit in a specific byte (held in byte array of nvm) to on (1) or off (0)

(tests>unit>lib/pysquared>test_bitflags.py) Implemented pytests for get() and toggle() functions (checking and setting bits in bytes of nvm byte array)

(nateinaction) New boot counter class (lib>pysquared>nvm>counter.py) and pytests (tests>unit>lib/pysquared>test_counter.py):
- get() -> int: Returns value of counter in nvm
- increment() -> None: Increments counter in nvm by one (now properly rolls over after reaching above 255)


nateinaction (commented):
General structure of this looks great. I made naming change suggestions in #98.

Also noticed that flag usages still expect class variables to have `__get__` `__set__` descriptors rather than the new methods you created. Will need to find and fix those before merge. 🚀

nateinaction (approved):

PR #96, opened 2025-01-17
Changes to hardware initialization

dbgen1:
Hardware initialization occurs in `pysquared.py`, where all the devices are initialized. Many of these devices the same behavior when it comes to their initialization, so it is beneficial to write functions that can handle initialization behavior. This PR does a couple things:

1. Add a `init_general_hardware` function which is able to handle the hardware initialization for any device that simply needs a firmware constructor. To keep the current code behavior unchanged, this function also optionally takes a callable which modifies the initialization behavior if `self.orpheus` is true. A detailed docstring in this function describes its behavior in more detail. 
2. Moves the initialization of hardware devices into their own functions, which are called in the constructor of `Satellite`. Using separate single-use functions for implementation makes the code self-documenting and allows for easier testability as per (3).
3. Add a `@safe_init` decorator to the `Satellite` class which wraps all the initialization functions in the standard try catch used for every hardware instance with a defined error severity if needed, and can print detailed debug information if `self.debug` is enabled.


Mikefly123:
Hey @dbgen1! The proposed changes sound like they could be pretty great. Particularly, the creation of individual functions for hardware initialization seems like a good abstraction to increase code readability and also probably allow us to individually deinit and reinit individual pieces of hardware more easily. 

I had a separate question though about this PR, it looks like there are a bunch of files that are "changed" but are identical to the origin files. Not super familiar with why this happens sometimes with GitHub. Is it a file formatting issue?

dbgen1:
I'm not sure why it happens. I noticed it as well and used a diff tool to verify that nothing  important was *actually* changed. It might have to do with the way I cloned the repo to my machine?

Mikefly123:
@nateinaction have you seen this happen before? I think it happened once or twice before to me on Mac. @dbgen1 are you using cloud syncing by any chance. I think this was happening to be when I was cloning my repos to a folder that was being backed up by my iCloud Drive and it went away once I started cloning to a folder that was not backed up by the iCloud. 

I think I fixed the issue by removing and then recloning the repository.

nateinaction (changes requested):
Hmmm. This is pretty strange and I would think that the pre-commit hook would have caught this. It looks like an extra invisible char was added to every line by your editor. I suspect some delving into your editor config might solve this problem but you also might need to reapply your changes onto a new branch:
```diff
git diff fc514a307c59cfbd83929da8eb3bbe78d1cca199 --word-diff-regex=.
diff --git a/.github/workflows/ci.yaml b/.github/workflows/ci.yaml
index db70d31..c1b8104 100644
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,53 +1,53 @@
name: ci{+^M+}
{+^M+}
on:{+^M+}
  pull_request:{+^M+}
  push:{+^M+}
    branches:{+^M+}
      - main{+^M+}
{+^M+}
jobs:{+^M+}
  lint:{+^M+}
...
```

Even though I've seen the code in office hours and like the idea, the PR isn't easily reviewable in this state. Blocking.

dbgen1:
The issue was caused by the editor changing the whitespace at the end of each line from a Unix style (LF) line ending to a windows style (CRLF) line ending. I configured git to automatically convert the line ending style to the correct one used in this repository (LF). The latest commit contains all the actual changes and shows no other modifications.

lib/pysquared/pysquared.py
```diff
@@ -88,10 +88,155 @@ def debug_print(self, statement: Any) -> None:
             print(co("[pysquared]" + str(statement), "green", "bold"))
 
     def error_print(self, statement: Any) -> None:
-        self.c_error_count += 1  # Limited to 255 errors
+        self.c_error_count: multiBitFlag = (
+            self.c_error_count + 1
+        ) & 0xFF  # Limited to 255 errors
```
nateinaction, line 93: Looks like inadvertent conflict resolution that removed #87

lib/pysquared/pysquared.py
```diff
@@ -202,6 +348,8 @@ def __init__(self) -> None:
         """
         NVM Parameter Resets
         """
+        if self.c_boot > 200:
+            self.c_boot = 0
```
nateinaction, line 352: Looks like inadvertent conflict resolution that removed #87

nateinaction (commented):
Much better! Looks like there was an issue with resolving a merge conflict that removed #87. Once that's fixed I'll test this on my board and give you a 👍

lib/pysquared/pysquared.py
dbgen1: This has been fixed.

dbgen1:
Going to do one more commit to appease the linter and this should be good to go!

lib/pysquared/pysquared.py
```diff
@@ -92,6 +92,149 @@ def error_print(self, statement: Any) -> None:
...
+
+            Returns:
+                Any: The initialized hardware instance if successful, or `None` if an error occurs.
+
+        Raises:
+                Exception: Any exception raised by the `init_func` or `orpheus_func`
+                will be caught and handled by the `@safe_init` decorator.
+        """
+        if self.orpheus and orpheus_func:
+            return orpheus_func(hardware_key)
```
nateinaction, line 144: It looks like this stops the hardware from being initialized but does not stop the hardware from being set to an initialized state in the hardware dictionary.

```sh
Device  | Status
|I2C0   | True |
```

dbgen1: I fixed this by moving the dictionary setting to each individual init function rather than the decorator, which is a method that makes more logical sense anyway.

nateinaction (approved):

PR #97, opened 2025-01-17
Update radio test to work with new repo structure

nateinaction:
This PR updates the radio test so that it can be easily imported into the repl. It also includes a readme on how to run it. The TX State team was able to successfully send jokes to running flight control board using this test file.

### DIFF
--- a/tests/repl/radio_test.py
+++ b/tests/repl/radio_test.py
@@ -259,9 +259,12 @@ class RadioTest:
 
         while True:
             if device_selection == "A":
+                self.number_of_attempts += 1
+                if self.number_of_attempts >= 5:
+                    print("Too many attempts. Quitting.")
+                    break
                 time.sleep(1)
                 self.device_under_test(self.number_of_attempts)
-                self.number_of_attempts += 1
             elif device_selection == "B":
                 time.sleep(1)
                 self.receiver()
--- a/tests/repl/radio_test.py
+++ b/tests/repl/radio_test.py
@@ -6,278 +6,265 @@
 
 import time
 
-from pysquared import cubesat
-
-test_message = "Hello There!"
-debug_mode = True
-number_of_attempts = 0
-cube_callsign = ""
-
-if cube_callsign == "":
-    print("No cube callsign!")
-    exit()
-
-# Radio Configuration Setup Here
-radio_cfg = {
-    "spreading_factor": 8,
-    "tx_power": 13,  # Set as a default that works for any radio
-    "node": 0x00,
-    "destination": 0x00,
-    "receive_timeout": 5,
-    "enable_crc": False,
-}
-
-if input("FSK or LoRa? [L/f]") == "F":
-    cubesat.f_fsk = True
-    del cubesat
-    print("Resetting in FSK")
-    from pysquared import cubesat
-
-print("FSK: " + str(cubesat.f_fsk))
-
-options = ["A", "B", "C"]
-
-# Setting the Radio
-cubesat.radio1.spreading_factor = radio_cfg["spreading_factor"]
-if cubesat.radio1.spreading_factor > 8:
-    cubesat.radio1.low_datarate_optimize = True
-else:
-    cubesat.radio1.low_datarate_optimize = False
-cubesat.radio1.tx_power = radio_cfg["tx_power"]
-cubesat.radio1.receive_timeout = radio_cfg["receive_timeout"]
-cubesat.radio1.enable_crc = False
-
-print(
-    """
-=======================================
-|                                     |
-|              WELCOME!               |
-|       Radio Test Version 1.0        |
-|                                     |
-=======================================
-|       Please Select Your Node       |
-| 'A': Device Under Test              |
-| 'B': Receiver                       |
-================ OR ===================
-|      Act as a client                |
-| 'C': for an active satalite         |
-=======================================
-"""
-)
+from lib.pysquared.pysquared import Satellite
 
 
-def debug_print(message):
-    if debug_mode:
-        print(message)
+class RadioTest:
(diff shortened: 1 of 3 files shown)


Mikefly123:
Reading through this it looks good to me! I don't have the hardware to test this right now, but I can try to get to it on Tuesday when I return to the lab.

nateinaction:
Yeah, this was kind of hacky. Just wanted to get something working.

I suspect the team will spend a lot more focused on actually testing these interactions. When we do, working and testing stuff in the repl will get easier.

tests/repl/radio_test.py
```diff
@@ -6,278 +6,265 @@
...
 
-def debug_print(message):
-    if debug_mode:
-        print(message)
+    def device_under_test(self, attempts):
+        self.debug_print("Device Under Test Selected")
+        self.debug_print("Setting up Radio...")
 
+        self.cubesat.radio1.node = 0xFA
+        self.cubesat.radio1.destination = 0xFB
```
Mikefly123, line 59: In order to correctly "ping-pong" when we are the `device_under_test` we should configure as:
```py
self.cubesat.radio1.node = 0xFB
self.cubesat.radio1.destination = 0xFA
```
This matches the way the we setup the standard flight code.

tests/repl/radio_test.py
```diff
@@ -6,278 +6,268 @@
...
-# Setting the Radio
-cubesat.radio1.spreading_factor = radio_cfg["spreading_factor"]
-if cubesat.radio1.spreading_factor > 8:
-    cubesat.radio1.low_datarate_optimize = True
-else:
-    cubesat.radio1.low_datarate_optimize = False
-cubesat.radio1.tx_power = radio_cfg["tx_power"]
-cubesat.radio1.receive_timeout = radio_cfg["receive_timeout"]
-cubesat.radio1.enable_crc = False
+        cubesat.f_fsk = True if fsk_or_lora == "fsk" else False
```
Mikefly123, line 31: I think using the ternary operator here is a little confusing. Also, the `fsk_or_lora == "fsk"` condition doesn't match the comment above which says `# Either "f" for FSK or "l" for LoRa`

tests/repl/radio_test.py
```diff
@@ -6,278 +6,268 @@
...
+            passcode = input(
+                "What's the passcode (in plain text, will automagically be converted to UTF-8): "
+            )
+
+        while True:
+            if device_selection == "A":
+                self.number_of_attempts += 1
+                if self.number_of_attempts >= 5:
+                    print("Too many attempts. Quitting.")
+                    break
```
Mikefly123, line 265: Breaking from being the transmitter after a certain number of attempts is not a bad idea, but a few thoughts if we intend to continue with this addition: 

- The standard practice for using `radio_test.py` right now is to just turn on the `device_under_test` function and let it run indefinitely while the test is being run. A test operator can simply interrupt the code when they are done. 
- If we want to auto end the test after certain number of tests, I would recommend making the test duration a dynamic input rather than just hard coding 5 attempts (quite a low number) as the only setting. 
- Dynamic input could be accomplished by just added another `input()` call before the test begins where the user can state how long they want the test to run for (bonus points if it can accept time as an input rather than just the number of tests! 

Just to keep the contribution history organized it might make sense to break this out into its own independent PR.

Mikefly123 (changes requested):

Mikefly123:
Hey @nateinaction I was wondering if we could revisit this and approve / merge the PR so we can get this functionality updated and keep this from getting too stale. 

I think from my requested changes on the `0xFA` and `0xFB` swap is critical

nateinaction:
@Mikefly123 Ok, this has been added to my short list. Will take a look at this hopefully tonight maybe tomorrow night.

tests/repl/radio_test.py
nateinaction: At this point the radio has already been instantiated (in the `Satellite` class [which happens on repl boot] that has been passed into the `RadioTest` class. Setting fsk here would only impact the next time the repl rebooted... would probably result in unexpected test behavior. Bringing the work from my radio branch could make this kind of fun.

Mikefly123: Hm, yeah that does raise an interesting challenge. Makes it more relevant than ever that we try to figure out a way to implement good `init` and `deinit` routines to allow for reconfiguring the hardware during the same runtime without needing a reboot!

nateinaction: I'm not sure I agree that we want that functionality. Using FPrime as a model we would only allocate memory once at application start. We wouldn't touch the memory model after that point. Kind of hard to do in Python but in that world we would prefer reboots into the state that we want rather than risk some kind of memory overrun on the running spacecraft. I can understand how that feature might help a ground station controller though.

Mikefly123: Hm you do make a good point. I think that the radio is a particularly tricky case in that switching between `fsk` and `LoRa` modes kind of requires tearing down a bunch of settings and building them back up. I think on the flight software side it is perfectly acceptable to just initiate a restart if you need to do a major reconfiguration of something like the radio. 

Why don't we just stick to a static default of being in LoRa mode for the `radio_test.py` that is in this repo for now. 

For runtime mode switching, maybe we could take what you have sketched out in #143 and file that into the new [ground_station_software repo](https://github.com/proveskit/ground-station-software)? It is currently empty, but I plan to add to it all the code needed to bring up a flight controller board as a ground station controller.

Mikefly123:
As mentioned in my review of #161 I think I will lay out a task for holistically changing how we do REPL testing in the near future. We might just want to close this PR for now and rebase on the incoming `radio_manager` system before opening this back up!

PR #100, opened 2025-01-19
Enforce lf line endings

nateinaction:
Doing our best to enforce `lf` line endings to prevent PRs with lots of accidental line changes. Thoughts @dbgen1?


dbgen1 (approved):
This is a great change for consistency. Nice touch enforcing in both the pre-commit and the linter! I tested all three measures (pre-commit, linter, vsc) on a windows system, and it successfully ensured LF line endings in each case. Approved ✅

PR #101, opened 2025-01-20
Add Config file test

yudataguy:
Since config is mission critical, add test for `config.json` and pre-commit json check. I think some of the field can have a range, like `NORMAL_MICRO_TEMP`, if there's suggestions for the value range, please let me know, so i can add them to the tests

### DIFF
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -27,7 +27,7 @@ CONFIG_SCHEMA = {
     "CRITICAL_BATTERY_VOLTAGE": float,
     "vlowbatt": float,
     "battery_voltage": float,
-    "current_draw": int,
+    "current_draw": float,
     "REBOOT_TIME": int,
     "turbo_clock": bool,
     "radio_cfg": dict,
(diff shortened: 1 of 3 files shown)


nateinaction (dismissed):
Sorry for the delayed review. Great contribution. Thank you!

Mikefly123 (approved):
Looks great to me! It will indeed be awesome to get more of this testing infrastructure built out. This is a solid start.

tests/unit/test_config.py
```diff
@@ -0,0 +1,258 @@
...
+    """Test that voltage values are within expected ranges."""
+    voltage_fields = [
+        "battery_voltage",
+        "NORMAL_BATTERY_VOLTAGE",
+        "CRITICAL_BATTERY_VOLTAGE",
+        "vlowbatt",
+    ]
+    for field in voltage_fields:
+        value = config_data[field]
+        assert 0 <= value <= 12.0, f"{field} must be between 0V and 12V"
```
Mikefly123, line 235: You could actually constrain this further to be only within the acceptable range of our battery packs. 
```suggestion
        assert 5.2 <= value <= 8.4, f"{field} must be between 5.2V and 8.4V"
```

Mikefly123 (approved):

PR #105, opened 2025-01-21
Use structured logging

blakejameson:
Implemented the beginning stages of a Logger class

The logger is used in replace of the debug_print and print statements.

Regarding the error_print statements, I have currently refactored the error_print function to use the logger internally. The error_print has an increment to self.c_error_count, so refactoring wasn't a simple switch to the logger error message and required some thought to approach a refactor. I am following Nate's guidance to refactor this section after his and Taylor's merge to the change counters (#93 #95).

This logger class currently has the functionality to format key, value pairs into a json string, which can be printed to standard output. 

Currently, there **IS NOT** the ability to save these logs to a file. This functionality is a longer-term objective.

I tested this code by running it on a v4b board, and the output shows similar to the main branch, the only difference of course seeing the output in json


blakejameson:
UPDATE:

Resolved error that was occurring after rebase. A function call was missing an argument.

Can confirm I ran the code on the board and had similar output to that of main branch

nateinaction (commented):
This was a lot of work and it's almost in! I like the logger interface that you created, that looks nice and easy to use. I noticed some repeated code in the logger implementation. What do you think of this? https://github.com/proveskit/circuitpy_flight_software/pull/109?

I know the old logger in this repo tracked file names and I played around with the concept of tracking filenames with the interface you created for this logger but I'm just not won over. Log messages are usually unique enough to help track down issues, tracking files adds boiler plate `filename=` vars at the top of every file that wants to use it, and I don't see this kind of tracking in other loggers so what about nixing it all together?

lib/requirements.txt
```diff
@@ -1,6 +1,8 @@
 adafruit-circuitpython-asyncio @ git+https://github.com/adafruit/adafruit_circuitpython_asyncio@1.3.3
 adafruit-circuitpython-drv2605==1.3.4
 adafruit-circuitpython-lis2mdl==2.1.23
+adafruit-circuitpython-logging==5.5.2
+adafruit-circuitpython-lsm303-accel==1.1.22
```
nateinaction, line 5: Looks like an accidental add. Removed in #109

blakejameson:
> This was a lot of work and it's almost in! I like the logger interface that you created, that looks nice and easy to use. I noticed some repeated code in the logger implementation. What do you think of this? #109?
> 
> I know the old logger in this repo tracked file names and I played around with the concept of tracking filenames with the interface you created for this logger but I'm just not won over. Log messages are usually unique enough to help track down issues, tracking files adds boiler plate `filename=` vars at the top of every file that wants to use it, and I don't see this kind of tracking in other loggers so what about nixing it all together?

Thank you for the feedback in #109 

Regarding the filenames, I am with you. It didn't feel quite right regarding the 'filename=' at the top of different files and felt kinda clumsy or awkward. I am definitely supportive of getting rid of the filename parameter

lib/pysquared/battery_helper.py
```diff
@@ -52,7 +55,7 @@ def _wait_for_ack(self):
             if self.uart.in_waiting:
                 byte = self.uart.read(1)
                 if self.debug_mode:
-                    print(f"ACK byte received: {byte}")
+                    self.logger.info(message=f"ACK byte received: {byte}")
```
nateinaction, line 58: There is nothing wrong with this but... now that `message` is the first argument, we could leave off the key here. That would make the call to the logger feel similar to `loguru` [link](https://loguru.readthedocs.io/en/stable/overview.html#ready-to-use-out-of-the-box-without-boilerplate). It's less work for consumers of the logger and they can choose in interact with kwargs if they need to.

What about something like this?
```suggestion
                    self.logger.info("ACK received", "ack"=byte)
```

lib/pysquared/battery_helper.py
```diff
@@ -89,8 +92,7 @@ def _read_message(self, timeout_ms=150):
                 if start_idx < end_idx:
                     return text[start_idx + 1 : end_idx]
         except Exception as e:
-            print(f"Error decoding message: {e}")
-            pass
+            self.logger.error(message=f"Error decoding message: {e}")
```
nateinaction, line 95:
```suggestion
            self.logger.error("Error decoding message", "err"=e)
```

lib/pysquared/battery_helper.py
```diff
@@ -275,14 +275,16 @@ def debug_timing(self):
...
-        print(f"Parse time: {parse_time:.2f}ms")
-        print(f"Total time: {total_time:.2f}ms")
-        print(f"Response: {response}")
+        self.logger.info(
+            send_time=f"{send_time:.2f}ms",
+            read_time=f"{read_time:.2f}ms",
+            parse_time=f"{parse_time:.2f}ms",
+            total_time=f"{total_time:.2f}ms",
+            response=f"{response}",
+        )
```
nateinaction, line 290: Missing a message. What about something like this?
```suggestion
        self.logger.info(
            "timing analysis",
            send_time=f"{send_time:.2f}ms",
            read_time=f"{read_time:.2f}ms",
            parse_time=f"{parse_time:.2f}ms",
            total_time=f"{total_time:.2f}ms",
            response=f"{response}",
        )
```

main.py
```diff
@@ -12,26 +12,25 @@
...
 import microcontroller
 
+import lib.pysquared.pysquared as pysquared
 from lib.pysquared.config import Config
-from lib.pysquared.pysquared import Satellite
+from lib.pysquared.logger import Logger
+
+logger = Logger()
+
+logger.info(message=" ", software_version="2.0.0", published_date="November 19, 2024")
```
nateinaction, line 21:
```suggestion
logger.info("Booting", software_version="2.0.0", published_date="November 19, 2024")
```

nateinaction (commented):
You made quick work of that review yesterday! Awesome! I've left a few comments and suggestions on calls to the logger that could be implemented in more places the themes for the call outs are:

- Remove the `message=` key from the first argument in each logger call
- Ensure that every call to the logger has a non-empty message
- Move all variables injected into message strings to use key/value pairs (kwargs) that you feed to the logger
- Add tests for all of the public functions you've created in `logger.py`

Because of the conflicts with this large change, I want to make sure this is the next thing that goes in if we can so thank you for the quick follow up.

What do you think of the proposed changes?

lib/pysquared/Field.py
```diff
@@ -4,42 +4,38 @@
 Authors: Nicole Maggard, Michael Pham, and Rachel Sarmiento
 """
 
-import traceback
```
nateinaction, line 7: Noting for the future that we talked about removing traceback and we couldn't find a reason to keep it. We tried both in repl and found that reporting back the exception var `e` gave the same output but maybe we missed something. 🤷

lib/pysquared/logger.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+
+import json
+import time
+
+# from adafruit_blinka import Enum
+
+# class DebugMode(Enum):
+# PRINTMODE = False
+# DEBUGMODE = True
```
nateinaction, line 13:
```suggestion
```

This will always be preserved in the PR history if you want to look back!

lib/pysquared/logger.py
```diff
@@ -0,0 +1,85 @@
...
+    #     """
+    #     self.logToFile = True
+    #     self.logToStandardOut = False
+
+    # def setLogToStdOut(self):
+    #     """
+    #     Set the logger to output logs to standard output.
+    #     """
+    #     self.logToStandardOut = True
+    #     self.logToFile = False
```
nateinaction, line 39:
```suggestion
```

This will always be preserved in the PR history if you want to look back!

nateinaction (commented):
Had a few final comments. What are your thoughts?

Not directed at you or this PR but oof @ that big try/except block in `cdh.py`. Cleaning that up could make logging nicer but that makes sense to do in another PR.

lib/pysquared/battery_helper.py
```diff
@@ -275,14 +275,17 @@ def debug_timing(self):
...
-        print(f"Parse time: {parse_time:.2f}ms")
-        print(f"Total time: {total_time:.2f}ms")
-        print(f"Response: {response}")
+        self.logger.info(
+            "timing analysis",
+            send_time=f"{send_time:.2f}ms",
+            read_time=f"{read_time:.2f}ms",
+            parse_time=f"{parse_time:.2f}ms",
+            total_time=f"{total_time:.2f}ms",
+            response={response},
```
nateinaction, line 290:
```suggestion
            response=response,
```

lib/pysquared/Field.py
```diff
@@ -4,42 +4,38 @@
...
-
 from lib.pysquared.debugcolor import co
+from lib.pysquared.logger import Logger
 
 
 class Field:
     def debug_print(self, statement):
         if self.debug:
             print(co("[Field]" + statement, "pink", "bold"))
```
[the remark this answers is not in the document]
blakejameson: yes lol

lib/pysquared/Field.py
```diff
@@ -4,42 +4,38 @@
...
             if self.cubesat.is_licensed:
-                self.debug_print("I am beaconing: " + str(msg))
-                print(
-                    "Message Success: "
-                    + str(self.cubesat.radio1.send(bytes(msg, "UTF-8")))
+                self.logger.info("I am beaconing", beacon=str(msg))
+                self.logger.info(
+                    "Message Success?",
+                    success=str(self.cubesat.radio1.send(bytes(msg, "UTF-8"))),
                 )
```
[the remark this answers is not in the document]
blakejameson: Screenshot 2025-01-24 at 10 13 44 PM

It appears so. I previously had tried to match the order of output is shown in the image, but I don't see why the two can't be combined into a single message

lib/pysquared/cdh.py
```diff
@@ -122,17 +134,17 @@ def FSK(cubesat) -> None:
 
     def joke_reply(self, cubesat) -> None:
         joke: str = random.choice(self._jokereply)
-        print(joke)
+        self.logger.info(joke)
```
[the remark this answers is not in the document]
blakejameson: Thank you for providing the "Sending joke reply" string for the message string. However, since the _log function has kwargs["msg"] = message, this means that an argument of 'msg' can not be passed, as it would lead to msg=joke being disregarded.

In this example, kwargs["msg"] would end up equaling "Sending joke reply" and the joke would be lost. All I simply have to do is rename the joke message to a different key

lib/pysquared/cdh.py
```diff
@@ -122,17 +134,17 @@ def FSK(cubesat) -> None:
...
 
-    def shutdown(cubesat, args) -> None:
+    def shutdown(self, cubesat, args) -> None:
         # make shutdown require yet another pass-code
         if args == b"\x0b\xfdI\xec":
-            print("valid shutdown command received")
+            self.logger.info("valid shutdown command received")
             # set shutdown NVM bit flag
-            cubesat.f_shtdwn: bool = True
+            cubesat.f_shtdwn = True
```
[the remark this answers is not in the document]
blakejameson: This seems spot on

lib/pysquared/cdh.py
```diff
@@ -154,13 +166,14 @@ def shutdown(cubesat, args) -> None:
                 monotonic_time=time.monotonic() + eval("1e" + str(_t))
             )  # default 1 day
             # set hot start flag right before sleeping
-            cubesat.f_hotstrt: bool = True
+            cubesat.f_hotstrt = True
```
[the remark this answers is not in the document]
blakejameson: Yes, looks like another example of me messing up a line from not handling the merge conflict well

lib/pysquared/packet_sender.py
```diff
@@ -54,35 +57,44 @@ def send_packet_with_retry(self, packet, seq_num):
 
         return False
 
-    def send_data(self, data, progress_interval=10):
+    def send_data(self, data, logger: Logger, progress_interval=10):
         """Send data with minimal progress updates"""
         packets = self.pm.pack_data(data)
         total_packets = len(packets)
-        print(f"Sending {total_packets} packets...")
+        logger.info("Sending packets...", num_packets=total_packets)
```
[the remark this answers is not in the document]
blakejameson: This honestly seems to be a place where I glanced over and forgot about changing. Seems that I did things right with packet_manager but somehow forgot to change things in packet_sender?

No excuses from me; I just simply missed this

lib/pysquared/functions.py
```diff
@@ -419,26 +415,22 @@ def do_detumble() -> None:
                             data[0][x] = 0.0
                     data[0] = tuple(data[0])
                     dipole = detumble.magnetorquer_dipole(data[1], data[0])
-                    self.debug_print("Dipole: " + str(dipole))
-                    self.send("Detumbling! Gyro, Mag: " + str(data))
+                    self.logger.debug("Detumbling", dipole=dipole)
+                    self.send("Detumbling! Gyro, Mag:", data=data)
```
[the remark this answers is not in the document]
blakejameson: dangit. Yes.

nateinaction (approved):
image


Great job!

PR #112, opened 2025-01-26
Improve logger class

blakejameson:
I removed the error_print() calls that were previously in pysquared.py. Error_count became a property of the Logger object. With the move to the Logger, it still remained as an nvm Counter object.

I added 2 new fields to the config.json, log_level and log_mode; I removed the debug field

I removed the debug parameter from Face, AllFaces, and Field. Being that the debut property or rather now the 'log_level' state is managed by the logger, there is no need for those classes to have a debug property and check for 'if debug', as that functionality is provided by the Logger

I can confirm I ran the code on a v4b board and saw similar output to what is currently in main, the only difference being the improved Logger error calls with cleaner output than previous.


Mikefly123:
Hey Blake! These changes look solid to me. I noticed that the new SonarCloud check is failing though, and I believe it's on a part of the code on main that was brought in by @dbgen1 in #96. 

image

Seems to be a pretty straightforward fix to me. Not sure if we want to capture it in this PR or make a unique patch for it on main.

blakejameson:
> Hey Blake! These changes look solid to me. I noticed that the new SonarCloud check is failing though, and I believe it's on a part of the code on main that was brought in by @dbgen1 in #96.
> 
> image
> 
> Seems to be a pretty straightforward fix to me. Not sure if we want to capture it in this PR or make a unique patch for it on main.

Appreciate you taking the time to look things over Michael

With the earlier commit, some pytests were failing due to me not using a mock for the byte array. With @nateinaction 's help, he guided me to steps I needed to take to address the pytest errors. Now the tests are passing,
but it turns out the adding self to line 67 wasn't sufficient for the SonarCloud analysis, as it now shows 6 resulting errors from my change. I will need to explore how to address this

blakejameson:
> > Hey Blake! These changes look solid to me. I noticed that the new SonarCloud check is failing though, and I believe it's on a part of the code on main that was brought in by @dbgen1 in #96.
> > image
> > Seems to be a pretty straightforward fix to me. Not sure if we want to capture it in this PR or make a unique patch for it on main.
> 
> Appreciate you taking the time to look things over Michael
> 
> With the earlier commit, some pytests were failing due to me not using a mock for the byte array. With @nateinaction 's help, he guided me to steps I needed to take to address the pytest errors. Now the tests are passing, but it turns out the adding self to line 67 wasn't sufficient for the SonarCloud analysis, as it now shows 6 resulting errors from my change. I will need to explore how to address this

@Mikefly123 @nateinaction 
I was just now able to get things passing the Sonar check just now by putting '@ staticmethod' above the safe_init function definition

blakejameson:
@nateinaction after rebasing 10 minutes ago, I tried running the code again on the board just to verify things still worked.

Screenshot 2025-01-28 at 12 53 12 PM

I got hit with the error above. Adding static_method above the safe_init function is how I resolved the Sonar errors a few days back. 

I tested the code on the FC board after adding static_method 2 days ago, but now I realize I obviously messed up by either dragging an older version of pysquared.py or being on the wrong branch. Anyways, dealing with the Sonar errors for this safe_init function will likely take some refactoring of the safe_init functions because using the 'self' or 'cls' keyword didn't fix things.

I am curious to know if you'd be open to reviewing this PR (with Sonar errors) and allowing the changes to be merged in on potential approval and if we could dedicate another issue towards fixing the Sonar errors resulting from the safe_init functions. Let me know what your thoughts are

blakejameson:
@nateinaction Apologies if this whole sequence is a lot to read through, but I went ahead and put back a parameter in the safe_init function of 'error_severity'. During my PR initially, I removed this parameter because it isn't needed. The logger outputs the severity so this string is unnecessary. However, for whatever reason, having this parameter in the function allows for no Sonar errors, though it marks it as an issue.

dbgen1:
Hi there, if you want to remove the error severity, there is actually no need for an additional layer to the decorator. The decorator currently consist of:
- a inner function "wrapper" which wraps the hardware initialization function
- a "decorator" which allows the wrapper to be applied to any function with the @ symbol,
- and a "decorator factory" (safe_init). The only purpose of the decorator factory is to allow arguments in a decorator, so if there is no error severity, this layer is not needed.

Theoretically, all the problems can be solved by simply removing the outer layer and renaming `decorator` to `safe_init`. Feel free to try it yourself, or if you are okay with it, I can commit a fix for this to this branch in a few hours!

blakejameson:
@Mikefly123 I am unable to see the Sonar reports before the most recent, but when I followed the recommendation of putting 'self', it resulted in 6 errors for the functions that used the decorator. I commented about it higher up in this thread

blakejameson:
@dbgen1 Thank you for explaining Davit. 
And if it's alright, I will let you commit the fix in a few hours. The world of decorators and wrappers is something I honestly haven't learned yet, so I am going to hold off in fear of still messing things up despite your excellent instructions

dbgen1:
@blakejameson Ready to go!

lib/pysquared/logger.py
```diff
@@ -6,10 +6,71 @@
...
+        self.datastore: ByteArray = datastore
+        self.levels_map: dict = {
+            "NOTSET": 0,
+            "DEBUG": 10,
+            "INFO": 20,
+            "WARNING": 30,
+            "ERROR": 40,
+            "CRITICAL": 50,
+        }
+        self.log_modes_set: set = {"PRINT", "FILE", "BOTH"}
```
nateinaction, line 40: Can we turn these into an ENUM or custom type instead of a string that we parse?

lib/pysquared/logger.py
```diff
@@ -3,15 +3,51 @@
 Logs can be output to standard output or saved to a file (functionality to be implemented).
 """
 
+# NVM register number
```
nateinaction, line 6:
```suggestion
```

lib/pysquared/pysquared.py
```diff
@@ -32,6 +31,7 @@
 from lib.pysquared.config import Config  # Configs
 from lib.pysquared.nvm.counter import Counter
 from lib.pysquared.nvm.flag import Flag
+from lib.pysquared.nvm.registers import BOOTCNT, FLAG
```
nateinaction, line 34: Sorry, I feel like this one is pedantic but let me know what you think...

What if the name of the file is `register` and we `import ... as register`? I feel like the rest of the code might read a little more easily if the registers are written as `register.FLAG`, etc. What do you think?

lib/pysquared/logger.py
```diff
@@ -3,15 +3,51 @@
...
+    ERROR = 4
+    CRITICAL = 5
+
 
 class Logger:
-    def __init__(self) -> None:
+    def __init__(
+        self,
+        datastore: ByteArray,
+        log_level: int = LogLevel.DEBUG,
```
nateinaction, line 38: Should this be?
```suggestion
        log_level: int = LogLevel.NOTSET,
```

lib/pysquared/logger.py
```diff
@@ -24,35 +60,39 @@ def _log(self, level: str, message: str, **kwargs) -> None:
 
         json_output = json.dumps(kwargs)
 
-        if self.logToStandardOut:
+        if self.logToStandardOut and self.can_print_this_level(level_value):
             print(json_output)
```
Mikefly123, line 64: If we so fancied we could return the debug color stuff here haha. Maybe something for a future PR

lib/pysquared/logger.py
```diff
@@ -3,15 +3,51 @@
...
-    def __init__(self) -> None:
+    def __init__(
+        self,
+        datastore: ByteArray,
+        log_level: int = LogLevel.DEBUG,
+        log_mode: int = LogMode.PRINT,
+    ) -> None:
+        self.datastore: ByteArray = datastore
         self.logToStandardOut: bool = True
+        self.error_count: Counter = Counter(index=ERRORCNT, datastore=self.datastore)
```
nateinaction, line 43: Ok, looking at this now. I think I might have given bad advice. Instantiating this class inside the logger constructor doesn't feel right and violates our desired state of having a dependency injection framework. I think the logger interface in `main` and `repl` might instead be better off looking like:
```py
logger: Logger = Logger(
    counter=Counter(index=ERRORCNT, datastore=self.datastore)
)
```

What do you think about trying to implement it that way?

lib/pysquared/pysquared.py
blakejameson: I really like this idea; great suggestion!

lib/pysquared/logger.py
blakejameson: eh, they act the same, but if it's preferred I can change it to NOTSET

lib/pysquared/logger.py
blakejameson: I am down for that!

main.py
```diff
@@ -16,20 +16,20 @@
 from lib.pysquared.config import Config
 from lib.pysquared.logger import Logger
 
-logger = Logger()
-
+logger: Logger = Logger(microcontroller.nvm)
 logger.info("Booting", software_version="2.0.0", published_date="November 19, 2024")
```
Mikefly123, line 20: This statement should actually probably be updated in each new PR to reflect this week's latest software version and the date that the PR was made ready for review. I think it's handy so users have an idea of what software they're running.

For this PR the version would be for `V2.0.0-alpha-25w05 Pre-Release`.

nateinaction: I think I can kind of solve for versioning in the build step soon!

lib/pysquared/logger.py
nateinaction: Yeah! We talked about using [the color interface from loguru](https://loguru.readthedocs.io/en/stable/overview.html#pretty-logging-with-colors) as a guide for how we might reimplement it in our new logger. 🚀

main.py
Mikefly123: Will be very excited to see that come in!

lib/pysquared/logger.py
Mikefly123: I don't have a finger on the scale in terms of design here, so up to you guys how to implement this! I did want to comment just to say this was the first time I had heard of a dependency injection framework, so I had to educate myself about it real quick. 

I can definitely see the merits and how this is building toward's Nate's vision of having much more modular code vs the doom stack that `pysquared.py` and its children have been. I think there's perhaps an interesting paper here to write about how flight software gets written when your team isn't traditionally trained in embedded software engineering. It feels like the approaches are going to end up being quite a bit different!

lib/pysquared/logger.py
```diff
@@ -3,15 +3,51 @@
...
 class Logger:
-    def __init__(self) -> None:
+    def __init__(
+        self,
+        datastore: ByteArray,
+        log_level: int = LogLevel.DEBUG,
+        log_mode: int = LogMode.PRINT,
+    ) -> None:
+        self.datastore: ByteArray = datastore
         self.logToStandardOut: bool = True
```
Mikefly123, line 42: A nitpick, but this is the only thing in this file that is camel case rather than snake case. 
```suggestion
        self.log_to_standard_out: bool = True
```

lib/pysquared/logger.py
```diff
@@ -6,12 +6,40 @@
 import json
 import time
 
+from lib.pysquared.nvm.counter import Counter
+
+
+class LogMode:
+    PRINT = 1
+    FILE = 2
+    BOTH = 3
```
nateinaction, line 15: Since `LogMode` is not fully implemented do you want to remove references to it in this PR?

lib/pysquared/logger.py
```diff
@@ -24,35 +52,39 @@ def _log(self, level: str, message: str, **kwargs) -> None:
...
 
     def critical(self, message: str, **kwargs) -> None:
         """
         Log a message with severity level CRITICAL.
         """
-        self._log("CRITICAL", message, **kwargs)
+        self._log("CRITICAL", 5, message, **kwargs)
+
+    def get_error_count(self) -> int:
+        return self.error_counter.get()
```
nateinaction, line 90: It looks like this isn't used anywhere but if we make the class member vars private then we will need this getter.

lib/pysquared/logger.py
nateinaction: @Mikefly123 Isn't that a cool idea? Maybe not so surprisingly that's exactly what FPrime does. In addition to many other cool things, it's is a dependency injection framework building all objects it will ever need on start.

lib/pysquared/logger.py
blakejameson: great idea

lib/pysquared/logger.py
```diff
@@ -6,12 +6,40 @@
...
+    def __init__(
+        self,
+        error_counter: Counter,
+        log_level: int = LogLevel.NOTSET,
+        log_mode: int = LogMode.PRINT,
+    ) -> None:
+        self.log_to_standard_out: bool = True
+        self.error_counter: Counter = error_counter
+        self.log_level: int = log_level
+        self.log_mode: int = log_mode
```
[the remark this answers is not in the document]
blakejameson: Thank you for this call out

lib/pysquared/logger.py
```diff
@@ -6,12 +6,31 @@
...
+    def __init__(
+        self,
+        error_counter: Counter,
+        log_level: int = LogLevel.NOTSET,
+    ) -> None:
+        self._error_counter: Counter = error_counter
+        self._log_level: int = log_level
 
-    def _log(self, level: str, message: str, **kwargs) -> None:
+    def can_print_this_level(self, level_value: int) -> bool:
```
nateinaction, line 30: This is also probably not intended to be called outside of this class. Let's make it internal.
```suggestion
    def _can_print_this_level(self, level_value: int) -> bool:
```

lib/pysquared/logger.py
```diff
@@ -24,35 +43,39 @@ def _log(self, level: str, message: str, **kwargs) -> None:
 
         json_output = json.dumps(kwargs)
 
-        if self.logToStandardOut:
+        if self.can_print_this_level(level_value):
```
nateinaction, line 46:
```suggestion
        if self._can_print_this_level(level_value):
```

nateinaction (approved):

PR #99, opened 2025-01-19
Improved Windows support in Makefile

nateinaction:
This PR ensures Makefile compatibility with a native Windows workflow. This PR also creates new development guides on how to get started working in the repo for MacOS, Linux, and Windows users.

Tested all make commands on Windows, Mac and Linux machines.

PR #113, opened 2025-01-27
Use UV for python env management

nateinaction:
Following #99, we found that a few Windows users had python setups that didn't allow the `make` commands to complete successfully. There are a few ways to fix this. The first method that I tried was to create a  [dev container](https://containers.dev/) but I didn't enjoy programming in the dev container. It took about 3 minutes to boot up anytime I started my IDE... yuck!

Next I tried UV and I think it makes a lot of our build simpler. What do you all think?


Mikefly123:
This sounds pretty cool! I don't have much to contribute to this, but I did want to link this pretty good explainer I watched about UV a while ago: 
https://youtu.be/gSKTfG1GXYQ?si=PGzudeVNXqxafxRw

Mikefly123:
I just gave this code a try on my Windows Machine and it is lightning quick! I thought for a while I was network throttled downloading packages, but I guess `pip` is perhaps past its prime. 

image

For comparison the "old" `make` process would take around 15s to download the libraries (~53s overall) vs around 3 s (~40s overall) with the new `uv` method. The big limiting factor is that copy time to get all the code loaded onto the RP2040. Maybe in the future it would be possible to check the version of the libraries currently on the RP2040 and only remove & update the ones that are out of date?

Mikefly123 (approved):
LGTM!

PR #114, opened 2025-01-27
Update config.json to Default is_licensed as False

Mikefly123:
Changed default for `is_licensed` from `True` to `False`

This blocks the radio from transmitting automatically when `main.py` is run. This avoids possible damage to the radio if there is no antenna attached and mitigates the risk that an illegal transmission is made if no licensed amateur radio operator is present.

### DIFF
--- a/config.json
+++ b/config.json
@@ -48,7 +48,7 @@
 "legacy": false,
 "heating": false,
 "orpheus":true,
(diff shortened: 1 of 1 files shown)


Alandlt15 (approved):
Great information, thank you @Mikefly123!

PR #115, opened 2025-01-27
Expand Windows Makefile Compatability

Mikefly123:
# What is Changed
Hey team! I'm on this fork on the main repo to try and work on some updates specifically for the InspireFly Version of the PROVES Kit that the Virginia Tech folks are flying. I noticed when I tried to do the `make` process on my windows machine that I kept running into permission errors. 

I think the permission errors were due to the fact that I had *both* a Windows Store download of Python and a direct install of Python on my machine at the same time. When I uninstalled the Windows Store Python, I would instead get a `Python was not found` error like the attached screenshot: 

image

I verified that Python was on the path and all that, but it seemed like the `Makefile` was expecting to find Python in the Windows Store install location even though it was no longer there. I make some changes to the `Makefile` to essentially check which Python is currently on the path and then use that for setting up the `venv` and it seems to work like a charm now! 

It would be great to get some extra feedback on how this works on other people's windows machines. I don't have a bunch of experience with `make`, so if this change breaks functionalities in an unexpected way please let me know!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,24 +9,26 @@ help: ## Display this help.
 
 ifeq ($(OS),Windows_NT)
 BIN_DIR := venv/Scripts
-PYTHON := $(BIN_DIR)/python.exe
+PYTHON := $(shell which python)
+VENV_PYTHON := $(BIN_DIR)/python.exe
 else
 BIN_DIR := venv/bin
-PYTHON := $(BIN_DIR)/python
+PYTHON := python3
+VENV_PYTHON := $(BIN_DIR)/python
 endif
 
 ACTIVATE_VENV := . $(BIN_DIR)/activate;
 
 venv:
 	@echo "Creating virtual environment..."
-	@python3 -m venv venv
-	@$(ACTIVATE_VENV) pip install --upgrade pip --quiet
-	@$(ACTIVATE_VENV) pip install --requirement requirements.txt --quiet
+	@$(PYTHON) -m venv venv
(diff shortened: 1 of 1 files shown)


Alandlt15:
I tested on my machine and it also works fine. I do have the Windows Store Python version installed running the make command on Windows (Git Bash).

nateinaction:
@Mikefly123 Can you try running #113. It's a different way to split this hair. The benefit of #113 is that it should be entirely independent of any python environment on the developer's machine.

Mikefly123:
Hey @nateinaction are there updated instructions for running #113? I couldn't spot them if that was the case, so I just tried the `make` commands which didn't work. 
image

nateinaction:
@Mikefly123 Hmmm that's odd! It looks like you're not running the code in #113. Your terminal is showing `black .` which was removed quite a while ago and #113 has `make install ...`.

There should be no changes to the `make` commands you can run in #113. Can you take another look and make sure you're running on the right branch when you get a chance?

Mikefly123:
> @Mikefly123 Hmmm that's odd! It looks like you're not running the code in #113. Your terminal is showing `black .` which was removed quite a while ago and #113 has `make install ...`.
> 
> There should be no changes to the `make` commands you can run in #113. Can you take another look and make sure you're running on the right branch when you get a chance?

I cleaned my local repo, tried again and it worked like a charm! Quite a bit faster than the old build process too wow!

I'd be happy to retract this PR if we think we'll just adopt the new system with UV in the near future.

nateinaction:
Opting for #113. Closing.

PR #116, opened 2025-01-27
fixed 'False' to 'false' in config.json

Alandlt15:
Capitalized false slipped by in config.json. Changed to lowercase.

### DIFF
--- a/config.json
+++ b/config.json
@@ -48,7 +48,7 @@
 "legacy": false,
 "heating": false,
 "orpheus":true,
-"is_licensed":False,
+"is_licensed":false,
 "NORMAL_TEMP": 20,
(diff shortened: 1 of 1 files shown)


Mikefly123 (approved):
Thanks for the catch! I was actually wondering how this worked since Booleans in Python are usually capitalized. I figure that the `config.py` class casts them to a Boolean when it reads the json?

Alandlt15:
No worries, It was like 8am when I reviewed lol my bad. 

And that is what its doing in the constructor, but if the json syntax is not correct, it will throw a parsing error.

blakejameson:
merged with Alan's approval

PR #117, opened 2025-01-28
Removing the log_face_data, log_error_data functions, and log functions

blakejameson:
Deleting the log_face_data and log_error_data functions, as they are not called anywhere. Also deleting the log function, as first of all it is not used. Second of all, log functionality to save to a file will eventually be supported in the Logger class, not as a standalone function inside another class

### DIFF
--- a/lib/pysquared/functions.py
+++ b/lib/pysquared/functions.py
@@ -347,24 +347,6 @@ class functions:
         pass
 
     """
-    Logging Functions
-    """
-
-    def log_face_data(self, data) -> None:
-        self.logger.debug("Logging Face Data")
-        try:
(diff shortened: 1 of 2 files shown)


Mikefly123:
LGTM! Honestly, I'm not really sure why I hadn't removed these functions earlier because they have indeed done nothing for a long time now.

Mikefly123 (approved):

blakejameson:
@Mikefly123 Thanks so much for taking the time to review!
And it seems that sometimes the longer functions stay in a codebase, the easier it can be to overlook them, so I definitely understand how easily one can just skip over a function and forget it exists at all!

Issue #102, opened 2025-01-20
Add Radio Config to config.json


Mikefly123:
# Proposed Change
We currently have a dictionary in `pysquared.py` that stores values for configuring the radio settings. We should try and extract these to be contained in `config.json` for better modularity and easier parameter management! 
```py
        self.radio_cfg: dict[str, float] = {
            "id": 0xFB,
            "gs": 0xFA,
            "freq": 437.4,
            "sf": 8,
            "bw": 125,
            "cr": 8,
            "pwr": 23,
            "st": 80000,
        }
```

Alandlt15:
@Mikefly123, do you happen to know what the full 'name' for these keys are? I'm trying to put as much 'human-readable' values in the json file and I'm a little bit confused on what gs, sf, bw, cr, pwr, and st mean.  

Additionally, for the commands in cdh.py, Nate and I talked about how that dictionary might be better protected inside of the class, rather than in the json file. Simply because we imagine that these commands will and should not be modifiable at run-time. Could this be the same for 'id' and 'gs'? Or any other one? 

Thank you.

Mikefly123:
@Alandlt15 oh yeah sorry about that! 
- `id`: The 1 byte ID value for the radio you are transmitting from. Also known as the `node` in the Radiohead Format
- `gs`: The 1 byte ID value for the radio you are transmitting to (the Ground Station). Also known as `destination` in the Radiohead Format
- `freq`: The radio frequency in Mhz (Megahertz)
- `sf`: The radio `spreading_factor`
- `bw`: The radio `bandwidth` in Khz (Kilohertz)
- `cr`: The radio `coding_rate`
- `pwr`: The `transmit power` for the radio.
- `st`: Genuinely have no idea what this one is, it is a legacy thing that we should probably investigation. I think it is a `Start Time` or something?

Mikefly123:
Re: The command keys in `cdh.py`

Yes I concur that those should probably be left hard coded rather than modifiable during runtime. I think we actually want to leave all of the other radio parameters open to change during runtime if possible, so in the future we can tweak the inflight radio performance as needed.

Alandlt15:
Got it got it. I appreciate the clarification!

PR #120, opened 2025-01-29
Added 'radio_cfg' dictionary to json file.

Alandlt15:
Solving issue #102. In which we aimed to extract and contain the values of 'radio_cfg' in the json file. 

Implemented the getter, setter, and saver functions for it. And adapted pysquared.py to fit this change.


Alandlt15:
****NOT READY FOR REVIEW****

Will take care of the errors tomorrow morning.

lib/pysquared/config.py
```diff
@@ -26,94 +26,111 @@ def __init__(self) -> None:
...
         """Sets the string value in the config dictionary
         Does not save value to disk, will not persist through reboots
         """
         self._config[key] = value
 
-    def setList(self, key: str, value: str) -> None:
+    def set_list(self, key: str, value: str) -> None:
         """Sets the string value from a list inside of the config dictionary
         Does not save value to disk, will not persist through reboots
         """
```
blakejameson, line 84: Great job Alan! All the code changes look great, and I went ahead and ran all the code on a v4b board, worked normal

My only comment is regarding the docstring for set_list. 
Using the word 'sets', it gives me the impression the value passed becomes the sole value for the list

Being that the value passed is added to the list at a particular key, I think modifying the doctoring to something like
"""Adds the string value to the list specified by the key inside of the config dictionary"""

makes it more comprehensible. Again, great work man!!

Alandlt15: Thank you for suggesting this. 

I also made a small change to the comments on the 'set_dict' too.

nateinaction:
Thank you for spending the time to make those config vars more readable. This looks really good! 🚀

PR #130, opened 2025-01-31
removing unused parameters in 3 functions

blakejameson:
I removed an unused parameter in 3 functions

2 functions had a Logger object as a parameter. At some point in time for the initial Logger implementation, I hadn't yet used dependency injection for the packet_sender class, so I passed a parameter of a Logger to 2 functions.

I just realized I had forgotten to delete the logger parameters. Being that there is a self.logger member, that is now used in the packet_sender class. My apologies for the mistake

I also realized that the exec_cmd function has a Satellite parameter that isn't used

### DIFF
--- a/lib/pysquared/cdh.py
+++ b/lib/pysquared/cdh.py
@@ -175,6 +175,6 @@ class CommandDataHandler:
 
         cubesat.radio1.send(data=str(eval(args)))
 
-    def exec_cmd(self, args) -> None:
+    def exec_cmd(self, cubesat: Satellite, args) -> None:
         self.logger.info("Executing command", args=args)
         exec(args)
--- a/lib/pysquared/cdh.py
+++ b/lib/pysquared/cdh.py
@@ -175,6 +175,6 @@ class CommandDataHandler:
 
         cubesat.radio1.send(data=str(eval(args)))
 
-    def exec_cmd(self, cubesat: Satellite, args) -> None:
+    def exec_cmd(self, args) -> None:
         self.logger.info("Executing command", args=args)
         exec(args)
--- a/lib/pysquared/functions.py
+++ b/lib/pysquared/functions.py
@@ -349,7 +349,7 @@ class functions:
 
     # Goal for torque is to make a control system
     # that will adjust position towards Earth based on Gyro data
-    def detumble(self, dur: int = 7, margin: float = 0.2, seq: int = 118) -> None:
+    def detumble(self, dur: int = 7) -> None:
         self.logger.debug("Detumbling")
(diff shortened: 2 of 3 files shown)


lib/pysquared/functions.py
```diff
@@ -349,7 +349,7 @@ def OTA(self) -> None:
 
     # Goal for torque is to make a control system
     # that will adjust position towards Earth based on Gyro data
-    def detumble(self, dur: int = 7, margin: float = 0.2, seq: int = 118) -> None:
```
Mikefly123, line 352: This removal is fine because we'll be relying on the new Columbia x Northeastern developed magnetorquers for future missions! 

For context though if you're curious these magic numbers being passed configured the motor drivers that were used in the past (are still installed) in the XY solar panel boards.

lib/pysquared/cdh.py
```diff
@@ -175,6 +175,6 @@ def query(self, cubesat: Satellite, args) -> None:
 
         cubesat.radio1.send(data=str(eval(args)))
 
-    def exec_cmd(self, cubesat: Satellite, args) -> None:
+    def exec_cmd(self, args) -> None:
```
Mikefly123, line 178: So the reason that the satellite object was originally passed into this function is to ensure that when we enter the scope of the `exec_cmd` function whatever is run inside the `exec()` block would have immediate access to the `satellite` object in case your arbitrary code it wanted to call anything from it. This is part of why `pysquared.py` ended up coming a doom stack class, because doing so made it convenient to be able to access any part of the satellite using an arbitrary string passed to this `exec_cmd`. 

I think are ways to solve having to pass the satellite object here, like declaring the satellite object as `global` on creation (which I believe is kinda icky in Python) or just taking away this ability and hoping that we properly implement a more proper the air update solution. 

I think we should actually do a little more testing before removing this or maybe someone who better understands Python can just override me with some better ideas lol.

Mikefly123 (changes requested):

lib/pysquared/cdh.py
blakejameson: @Mikefly123 I greatly appreciate your explanation of the reason for passing the Satellite object! I will go revert this change then push

blakejameson:
@Mikefly123 Just bumping the re-request for review if you get a chance! I reverted the changes made to exec_cmd so all the only changes would be the Logger no longer passed in packet_sender and the 2 unused variables in functions's detumble()

Mikefly123 (approved):
LTGM! Sorry I didn't notice the changes were made!

blakejameson:
@Mikefly123 All good! Thanks for taking the time to review again

PR #136, opened 2025-02-03
POC: Radio abstraction from Satellite class

nateinaction:
There's way too much in here so will be rewriting into a new PR however I wanted to post the idea for the radio instantiation refactor. This POC extracts radio instantiation from the `Satellite` class and places it into its own module and builds the appropriate version (fsk vs lora) using the factory method. More thought should be put into what the satellite failure mode should look like for each hardware module.


nateinaction:
Opting for #140. Closing

PR #138, opened 2025-02-04
Improving cyclomatic complexity

blakejameson:
## Summary
Put in effort to help improve cyclomatic complexity, which deals with the number of paths through the functions. Essentially, I tried to reduce the amount of nesting where possible and enable a developer to better interpret functions. In some functions, refactoring resulted in a big improvement while others was a smaller amount. Ultimately, the amount of nesting and ifs one has comfort in working with is subjective, and you may not agree with some of the proposed solutions I offer. Please request changes in functions or comment where you think the previous implementation of the function is better as is than with my proposed change.

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)

Screenshot 2025-02-03 at 8 02 01 PM
Here is the output of running the code on a v4b board. Similar output to main branch

### DIFF
--- a/lib/pysquared/Field.py
+++ b/lib/pysquared/Field.py
@@ -20,14 +20,12 @@ class Field:
             return
 
         try:
-            self.logger.info(
-                "I am beaconing",
-                beacon=str(msg),
-                success=str(self.cubesat.radio1.send(bytes(msg, "UTF-8"))),
-            )
-
+            sent = self.cubesat.radio1.send(bytes(msg, "UTF-8"))
         except Exception as e:
             self.logger.error("There was an error while Beaconing", err=e)
+            return
+
+        self.logger.info("I am beaconing", beacon=str(msg), success=str(sent))
 
     def troubleshooting(self):
         # this is for troubleshooting comms
--- a/lib/pysquared/Field.py
+++ b/lib/pysquared/Field.py
@@ -13,13 +13,12 @@ class Field:
         self.logger = logger
 
     def Beacon(self, msg):
+        if not self.cubesat.is_licensed:
+            self.logger.debug(
+                "Please toggle licensed variable in code once you obtain an amateur radio license",
+            )
+            return
         try:
-            if not self.cubesat.is_licensed:
-                self.logger.debug(
-                    "Please toggle licensed variable in code once you obtain an amateur radio license",
-                )
-                return
-
             self.logger.info(
                 "I am beaconing",
                 beacon=str(msg),
--- a/lib/pysquared/Field.py
+++ b/lib/pysquared/Field.py
@@ -18,6 +18,7 @@ class Field:
                 "Please toggle licensed variable in code once you obtain an amateur radio license",
             )
             return
+
         try:
             self.logger.info(
                 "I am beaconing",
--- a/lib/pysquared/Field.py
+++ b/lib/pysquared/Field.py
@@ -14,16 +14,18 @@ class Field:
 
     def Beacon(self, msg):
         try:
-            if self.cubesat.is_licensed:
-                self.logger.info(
-                    "I am beaconing",
-                    beacon=str(msg),
-                    success=str(self.cubesat.radio1.send(bytes(msg, "UTF-8"))),
-                )
-            else:
+            if not self.cubesat.is_licensed:
                 self.logger.debug(
                     "Please toggle licensed variable in code once you obtain an amateur radio license",
                 )
+                return
+
+            self.logger.info(
+                "I am beaconing",
+                beacon=str(msg),
+                success=str(self.cubesat.radio1.send(bytes(msg, "UTF-8"))),
+            )
+
         except Exception as e:
             self.logger.error("There was an error while Beaconing", err=e)
 
--- a/lib/pysquared/functions.py
+++ b/lib/pysquared/functions.py
@@ -263,13 +263,10 @@ class functions:
             self.logger.debug("Listening")
             self.cubesat.radio1.receive_timeout = 10
             received = self.cubesat.radio1.receive(keep_listening=True)
-            if received is not None and "HAHAHAHAHA!" in received:
-                return True
-            else:
-                return False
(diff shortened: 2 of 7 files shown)


blakejameson:
^ just rebased onto the newest version of main from 30 minutes ago

lib/pysquared/Field.py
```diff
@@ -13,17 +13,19 @@ def __init__(self, cubesat, logger: Logger):
...
-                )
-            else:
-                self.logger.debug(
-                    "Please toggle licensed variable in code once you obtain an amateur radio license",
-                )
+            self.logger.info(
+                "I am beaconing",
+                beacon=str(msg),
+                success=str(self.cubesat.radio1.send(bytes(msg, "UTF-8"))),
+            )
```
nateinaction, line 27: Could make this try/except even more clear with something like
```py
try:
    sent = self.cubesat.radio1.send(bytes(msg, "UTF-8"))
except Exception as e:
    self.logger.error("There was an error while Beaconing", err=e)

if sent:
    self.logger.info("I am beaconing", beacon=str(msg))
```

or maybe

```py
try:
    sent = self.cubesat.radio1.send(bytes(msg, "UTF-8"))
except Exception as e:
    self.logger.error("There was an error while Beaconing", err=e)
    return

self.logger.info("I am beaconing", beacon=str(msg))
```

Thoughts?

lib/pysquared/functions.py
```diff
@@ -263,13 +263,10 @@ def listen_joke(self) -> bool:
             self.logger.debug("Listening")
             self.cubesat.radio1.receive_timeout = 10
             received = self.cubesat.radio1.receive(keep_listening=True)
-            if received is not None and "HAHAHAHAHA!" in received:
-                return True
-            else:
-                return False
+            return received is not None and "HAHAHAHAHA!" in received
```
nateinaction, line 266: Yessss...

lib/pysquared/pysquared.py
```diff
@@ -517,17 +517,19 @@ def RGB(self) -> tuple[int, int, int]:
...
+
+        # NEOPIX is initialized
+        try:
+            self.neopixel[0] = value
+        except Exception as e:
+            self.logger.error(
+                "There was an error trying to set the new RGB value",
+                err=e,
+                value=value,
+            )
```
nateinaction, line 532: Interesting that we need a try/except at all here since it looks like we're just setting the a value on the first element of a list. The thing is, objects in python don't always behave as you expect. Whatever is in that list could be something that can raise an exception... hmmm let's find out...

I think we're dealing with the `__set_item__` [descriptor](https://docs.python.org/3/howto/descriptor.html) of a [PixelBuf](https://docs.circuitpython.org/en/latest/shared-bindings/adafruit_pixelbuf/index.html#adafruit_pixelbuf.PixelBuf). At first glance, the [implemenation](https://github.com/adafruit/Adafruit_CircuitPython_Pixelbuf/blob/f21b005facc40d230c6462799573822563842f83/adafruit_pixelbuf.py#L300-L313) doesn't appear to be able to raise an exception so long as [the index exists](https://github.com/adafruit/Adafruit_CircuitPython_Pixelbuf/blob/f21b005facc40d230c6462799573822563842f83/adafruit_pixelbuf.py#L274). Digging further I see [some additional](https://github.com/adafruit/Adafruit_Blinka/blob/139335185b5e50fd28db618629b1c2c7e192f14c/src/adafruit_blinka/microcontroller/rp2040_u2if/rp2040_u2if.py#L410) possibilities for exceptions when the driver tries to write the values to the hardware.

Ok, I was going to suggest removing the try/except in exchange for a check to see if that index exists but this feels cleaner.

lib/pysquared/pysquared.py
```diff
@@ -517,17 +517,19 @@ def RGB(self) -> tuple[int, int, int]:
...
-                )
-        else:
+        if not self.hardware["NEOPIX"]:
             self.logger.warning("The NEOPIXEL device is not initialized")
+            return
+
+        # NEOPIX is initialized
+        try:
+            self.neopixel[0] = value
+        except Exception as e:
```
nateinaction, line 527: Interesting that we need a try/except at all here since it looks like we're just setting the a value on the first element of a list. The thing is, objects in python don't always behave as you expect. Whatever is in that list could be something that can raise an exception... hmmm let's find out...

I think we're dealing with the `__set_item__` [descriptor](https://docs.python.org/3/howto/descriptor.html) of a [PixelBuf](https://docs.circuitpython.org/en/latest/shared-bindings/adafruit_pixelbuf/index.html#adafruit_pixelbuf.PixelBuf). At first glance, the [implemenation](https://github.com/adafruit/Adafruit_CircuitPython_Pixelbuf/blob/f21b005facc40d230c6462799573822563842f83/adafruit_pixelbuf.py#L300-L313) doesn't appear to be able to raise an exception so long as [the index exists](https://github.com/adafruit/Adafruit_CircuitPython_Pixelbuf/blob/f21b005facc40d230c6462799573822563842f83/adafruit_pixelbuf.py#L274). Digging further I see [some additional](https://github.com/adafruit/Adafruit_Blinka/blob/139335185b5e50fd28db618629b1c2c7e192f14c/src/adafruit_blinka/microcontroller/rp2040_u2if/rp2040_u2if.py#L410) possibilities for exceptions when the driver tries to write the values to the hardware.

Ok, I was going to suggest removing the try/except in exchange for a check to see if that index exists but this feels cleaner.

nateinaction (commented):
Ok, only one really minor actionable comment on `Field.py`

lib/pysquared/Field.py
blakejameson: I really like the second for some reason; thanks for the great suggestion!

lib/pysquared/pysquared.py
blakejameson: Thank you for putting your thoughts out there and supplying these links. Greatly appreciated Nate

nateinaction (approved):

PR #144, opened 2025-02-05
Type hinting another round

blakejameson:
## Summary
Added type hinting to more functions in Big_Data, Field, packet_sender, packet_manager, functions 

I ran the code on a v4b board and it was the same output as the main branch. 
It is worth mentioning that type hinting isn't enforced in Python, so even a bad job of type hinting still won't affect the code running.

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)

### DIFF
--- a/lib/pysquared/Big_Data.py
+++ b/lib/pysquared/Big_Data.py
@@ -13,7 +13,7 @@ except Exception:
 class Face:
     def __init__(
         self, Add: int, Pos: str, tca: adafruit_tca9548a.TCA9548A, logger: Logger
-    ):
+    ) -> None:
         self.tca: adafruit_tca9548a.TCA9548A = tca
         self.address: int = Add
         self.position: str = Pos
--- a/lib/pysquared/Big_Data.py
+++ b/lib/pysquared/Big_Data.py
@@ -39,7 +39,7 @@ class Face:
         self.veml = None
         self.drv = None
 
-    def Sensorinit(self, senlist, address):
+    def Sensorinit(self, senlist, address) -> None:
         gc.collect()  # Force garbage collection before initializing sensors
 
         if "MCP" in senlist:
@@ -79,7 +79,7 @@ class Face:
 
 
 class AllFaces:
-    def __init__(self, tca: adafruit_tca9548a.TCA9548A, logger: Logger):
+    def __init__(self, tca: adafruit_tca9548a.TCA9548A, logger: Logger) -> None:
         self.tca: adafruit_tca9548a.TCA9548A = tca
         self.faces: list[Face] = []
         self.logger: Logger = logger
@@ -98,7 +98,7 @@ class AllFaces:
             self.faces.append(face)
             gc.collect()  # Clean up after each face initialization
 
-    def Face_Test_All(self):
+    def Face_Test_All(self) -> list[list[float]]:
         results: list[list[float]] = []
         for face in self.faces:
             if face:
--- a/lib/pysquared/Big_Data.py
+++ b/lib/pysquared/Big_Data.py
@@ -1,29 +1,38 @@
 import gc
 
+import lib.adafruit_tca9548a as adafruit_tca9548a  # I2C Multiplexer
 from lib.pysquared.logger import Logger
+
+try:
+    from typing import Union
+
+except Exception:
+    pass
 
 
 class Face:
-    def __init__(self, Add, Pos, tca, logger: Logger):
-        self.tca = tca
-        self.address = Add
-        self.position = Pos
-        self.logger = logger
+    def __init__(
+        self, Add: int, Pos: str, tca: adafruit_tca9548a.TCA9548A, logger: Logger
+    ):
+        self.tca: adafruit_tca9548a.TCA9548A = tca
+        self.address: int = Add
+        self.position: str = Pos
+        self.logger: Logger = logger
 
         # Use tuple instead of list for immutable data
-        self.senlist = ()
+        self.senlist: tuple = ()
         # Define sensors based on position using a dictionary lookup instead of if-elif chain
-        sensor_map = {
+        sensor_map: dict[str, tuple[str, ...]] = {
             "x+": ("MCP", "VEML", "DRV"),
             "x-": ("MCP", "VEML"),
             "y+": ("MCP", "VEML", "DRV"),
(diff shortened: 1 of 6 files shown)


lib/pysquared/Big_Data.py
```diff
@@ -1,29 +1,38 @@
...
 
+import lib.adafruit_tca9548a as adafruit_tca9548a  # I2C Multiplexer
 from lib.pysquared.logger import Logger
 
+try:
+    from typing import Union
+
+except Exception:
+    pass
+
```
Mikefly123, line 11: Is there a reason that we need to wrap the `Union` import in a try catch block?

lib/pysquared/Big_Data.py
```diff
@@ -64,26 +79,36 @@ def Sensorinit(self, senlist, address):
...
+            ("z-", 4),
+        ]
         for pos, addr in positions:
-            face = Face(addr, pos, tca, self.logger)
+            face: Face = Face(addr, pos, tca, self.logger)
             face.Sensorinit(face.senlist, face.address)
             self.faces.append(face)
             gc.collect()  # Clean up after each face initialization
 
     def Face_Test_All(self):
```
Mikefly123, line 101: Should we add a hint here that it returns a nested list?

lib/pysquared/cdh.py
```diff
@@ -98,7 +105,7 @@ def message_handler(self, cubesat: Satellite, msg) -> None:
                 if multi_msg:
                     # TODO check for optional radio config
                     self.logger.info("multi-message mode enabled")
-                response = cubesat.radio1.receive(
+                response: Any = cubesat.radio1.receive(
```
Mikefly123, line 108: I think the `cubesat.radio1.receive()` only returns strings. We should check the `adafruit_rfm` library to be sure.

lib/pysquared/cdh.py
```diff
@@ -161,23 +168,23 @@ def shutdown(self, cubesat: Satellite, args) -> None:
...
         alarm.exit_and_deep_sleep_until_alarms(time_alarm)
 
-    def query(self, cubesat: Satellite, args) -> None:
+    def query(self, cubesat: Satellite, args: Any) -> None:
         self.logger.info("Sending query with args", args=args)
 
         cubesat.radio1.send(data=str(eval(args)))
 
-    def exec_cmd(self, cubesat: Satellite, args) -> None:
+    def exec_cmd(self, cubesat: Satellite, args: Any) -> None:
```
Mikefly123, line 188: I think the args that are passed into `exec_cmd` can only be a string because the `exec` function doesn't attempt to cast it to a string.

lib/pysquared/functions.py
```diff
@@ -241,7 +241,7 @@ def listen(self) -> bool:
         try:
             self.logger.debug("Listening")
             self.cubesat.radio1.receive_timeout = 10
-            received = self.cubesat.radio1.receive_with_ack(keep_listening=True)
+            received: Any = self.cubesat.radio1.receive_with_ack(keep_listening=True)
```
Mikefly123, line 244: Same comment in this file as before, I think `receive_with_ack()` only returns strings.

lib/pysquared/functions.py
```diff
@@ -320,13 +320,9 @@ def get_battery_data(
...
         self,
-    ) -> List[
-        tuple[float, float, float],
-        tuple[float, float, float],
-        tuple[float, float, float],
-    ]:
+    ) -> List[tuple[float, float, float]]:
         try:
-            data: list = []
+            data: list[tuple[float, float, float]] = []
```
Mikefly123, line 325: I think this is actually even more gross because it is a list with three tuples inside of it (one tuple for each call).

lib/pysquared/Big_Data.py
blakejameson: Yes there is. The reason is that the 'typing' module isn't supported in CircuitPython, so without the try catch, the code would result in an Exception. Since typing is used for the developer experience in an IDE and being that types aren't enforced in Python, this is why the exception is caught and simply passed when it is run on the board.

See the 'CircuitPython-Specific Considerations' section and the code snippet below it in [CircuitPython Typing](https://learn.adafruit.com/creating-and-sharing-a-circuitpython-library/typing-information) as a reference

lib/pysquared/cdh.py
blakejameson: Honestly, this was a bit tough for me and I was confused along the way. With the 2 attachments below, this is how things appeared in VSCode when looking at the function signatures. It showed that receive was assigned to be the result of asyncio_to_blocking(asyncio_receive). Looking at all this was a bit overwhelming and confusing.

Screenshot 2025-02-05 at 1 22 43 PM
Screenshot 2025-02-05 at 1 29 07 PM 

All that being said, [RFM9x.receive()](https://docs.circuitpython.org/projects/rfm9x/en/latest/api.html#adafruit_rfm9x.RFM9x.receive) says that receive returns a bytearray or None

I am confused lol

lib/pysquared/Big_Data.py
blakejameson: Ahh, yes. Great catch! Disappointed that I somehow missed Face_Test_All lol

lib/pysquared/functions.py
blakejameson: Snap, you are right. Thank you for pointing this out, and yeah, that function signature looks so nasty

lib/pysquared/Big_Data.py
Mikefly123: Interesting! Thanks for the context

Mikefly123:
Hey @blakejameson! I saw your ping for re-review after the rebase. Were all of the comments from before addressed yet or are they still pending?

blakejameson:
@Mikefly123 When I pinged you, I thought they were addressed but I just realized I accidentally missed 1 of your comments, so I will go make that fix real quick. My apologies. I will @ you whenever I am ready again

blakejameson:
@Mikefly123 I am sorry for that but am ready now. It turns out I had made some edits once and stashed them because I wanted to run the code from Nikko's branch. And then I never popped the stash after returning to my branch, so those edits didn't come back. Just an epic fail by me

Mikefly123 (approved):
LGTM!

PR #148, opened 2025-02-06
Reimplement log color

Nikko-Adrian-Pacleb:
## Summary
Using co from debugcolor, a dictionary is made for each log type
```py
LogColors = {
    "NOTSET": "NOTSET",
    "DEBUG": co(msg="DEBUG", color="blue"),
    "INFO": "INFO",
    "WARNING": co(msg="WARNING", color="orange"),
    "ERROR": co(msg="ERROR", color="pink"),
    "CRITICAL": co(msg="CRITICAL", color="red"),
}
```
The colored output is made by finding the level string in the json and replacing it with the colored format.
```py
colored_json_output = (
                json_output.replace(
                    f'"level": "{level}"', f'"level": "{LogColors[level]}"'
                )
                if self.colorized
                else json_output
            )
            print(colored_json_output)
```
colorized argument is added and set to false by default
```py
class Logger:
    def __init__(
        self,
        error_counter: Counter,
        log_level: int = LogLevel.NOTSET,
        colorized: bool = False,
    ) -> None:
        self._error_counter: Counter = error_counter
        self._log_level: int = log_level
        self.colorized: bool = colorized
```

## How was this tested
- [x] Added new unit tests
- [ ] Other (Please describe)


lib/pysquared/logger.py
```diff
@@ -23,9 +33,11 @@ def __init__(
         self,
         error_counter: Counter,
         log_level: int = LogLevel.NOTSET,
+        colorized: bool = True,
```
nateinaction, line 36: Let's leave this off by default and let the users choose to turn it on if it's a feature they want. You should go ahead and pass `True` in the `main.py` and `repl.py` to maintain the behavior you're looking for. Eventually we'll need to turn it off before flight or logging to the disk so a `False` default feels right for this project. What do you think?

nateinaction (commented):
This looks pretty cool! Can you add unit tests for this?

Also I noticed that the acceptance criteria for ordering the output wasn't included here. That would make the logging output quite a bit more readable. Is that something you want to work on?

lib/pysquared/logger.py
Nikko-Adrian-Pacleb: This makes more sense since there is no one looking for the colored json output once its out and if we want to color it again, we can do it in the ground station. I changed the default value to False. I also set the logger colorized value in main and repl to True for now.

Nikko-Adrian-Pacleb:
I would love to try to add a unit test and do the other work that you mentioned but I do not know how to start on the unit test or what type of test is needed for this one. I am checking the other files for reference. Is there something you could recommend for me to check out?

Also I am a little confused by the acceptance criteria. Can you expand that for me?

nateinaction:
@Nikko-Adrian-Pacleb learning to write tests is pretty challenging but there are [some good videos](https://www.youtube.com/watch?v=YbpKMIUjvK8) about it. We can also chat more about it in an office hours session but I'll be unavailable most of the week so don't wait for me to try it out! The flight-software discord might be a good place to ask for this kind of help too!

Here is the location of the unit tests for logger: https://github.com/proveskit/circuitpy_flight_software/blob/5cc6c9e1a2759d56257495fa551973424c936072/tests/unit/lib/pysquared/test_logger.py

The remaining request in the acceptance criteria is to order the fields in the structured output, making timestamp first, level second, message third, and the rest can remain unordered. Your final output should look something like the image at the bottom of the ticket.

Screenshot 2025-02-04 at 22 08 23

Nikko-Adrian-Pacleb:
@nateinaction
For the unit test, since I used the co function from debugcolor.py do we need to have a unit test for that one first to make sure that it is working. There is an option of brute forcing text using the [ANSI](https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797) code but I don't think that is a good idea. 

For the json format, one option I did is change the order for when the other arguments are added
```py
        now = time.localtime()
        asctime = f"{now.tm_year}-{now.tm_mon:02d}-{now.tm_mday:02d} {now.tm_hour:02d}:{now.tm_min:02d}:{now.tm_sec:02d}"
        kwargs["time"] = asctime

        kwargs["level"] = level

        kwargs["msg"] = message
```
image

The problem I think that could arise here is that when kwargs has initial arguments, it will print those first then time, level, and message.

One option I could think of is create another object then add kwargs at the end
This code is not yet added nor tested: 
```py
        json_content = {}

        now = time.localtime()
        asctime = f"{now.tm_year}-{now.tm_mon:02d}-{now.tm_mday:02d} {now.tm_hour:02d}:{now.tm_min:02d}:{now.tm_sec:02d}"
        json_content["time"] = asctime

        json_content["level"] = level

        json_content["msg"] = message

        json_content["args"] = kwargs

        json_output = json.dumps(json_content)
```
As of now, I still haven't set up my board so I cannot test it but this is one option that I could think of.

blakejameson:
Hey @Nikko-Adrian-Pacleb, I really like your thinking regarding the consideration of the order in adding to the dictionary

However, something to note is that prior to Python 3.7*, order wasn't guaranteed for dictionaries, so I don't think the solution here will lead to your desired result. I think that usage of a collection like an OrderedDict could be of great use here. There is an example of an OrderedDict used in `self.hardware` in pysquared.py as reference.

*I mention prior to Python 3.7 because CircuitPython is based on Python 3.4

Nikko-Adrian-Pacleb:
@blakejameson thanks for letting me know about that. I did read about OrderedDict but I got confused on how to use it since I was initially planning to use kwargs for json. I forgot that I could just make another dictionary that has the format we need. I will try this out later. 

I'm not sure who to ask for this but same question for the testing. Do I first need to have a unit test for the co function since that function is used for the implementation of the color?

nateinaction:
@Nikko-Adrian-Pacleb Good question. You don't _need_ to add tests for `co()` but it's a good idea. And since `co()` is only used in `logger.py` it might be helpful to bring the functionality into the `logger.py` and delete `debugcolor.py`. Thoughts? 🚀

Nikko-Adrian-Pacleb:
@nateinaction Got you! thank you so much for the reply. I think it is good to place co at the logger file. Will implement it as soon as possible.

Nikko-Adrian-Pacleb:
fixed the order of the logs using OrderedDict as recommended and appended the kwargs object using the update method
```py
        now = time.localtime()
        asctime = f"{now.tm_year}-{now.tm_mon:02d}-{now.tm_mday:02d} {now.tm_hour:02d}:{now.tm_min:02d}:{now.tm_sec:02d}"

        json_order: OrderedDict[str, str] = OrderedDict(
            [("time", asctime), ("level", level), ("msg", message)]
        )
        json_order.update(kwargs)

        json_output = json.dumps(json_order)
```
This is the output:
image

Also I feel like there is no point in making colored_json_output variable and just overwrite json_output directly.
```py
            colored_json_output = (
                json_output.replace(
                    f'"level": "{level}"', f'"level": "{LogColors[level]}"'
                )
                if self.colorized
                else json_output
            )
            print(colored_json_output)
```
Should I change it? Or should I make 2 variables instead just in case we want to send the json output somewhere and still use the colored output for visuals. I am not sure how and when this will be used
```py
            colored_json_output = json_output.replace(
                f'"level": "{level}"', f'"level": "{LogColors[level]}"'
            )
            print(colored_json_output if self.colorized else json_output)
```

blakejameson:
@Nikko-Adrian-Pacleb Once, @nateinaction shared with me that we should just save normal json (without color) whenever we save the output to logs, so I think it's best to keep both around

blakejameson:
@Nikko-Adrian-Pacleb Great work! And the general convention is that the unit tests would be part of this issue/PR.

As a reminder, to test the unit tests, `make test` is your friend here and will test the unit test files

Nikko-Adrian-Pacleb:
The unit tests are added. According to @Mikefly123, the timestamp is being worked on so the unit test for testing the order of the string is not tested.

lib/pysquared/logger.py
```diff
@@ -7,9 +7,41 @@
 import time
 import traceback
 
+# from lib.pysquared.debugcolor import color
```
blakejameson, line 10: Now that color lives in this file, could you delete this comment?

lib/pysquared/logger.py
```diff
@@ -7,9 +7,41 @@
...
+from collections import OrderedDict
+
 from lib.pysquared.nvm.counter import Counter
 
 
+def color(msg, color="gray", fmt="normal"):
```
blakejameson, line 16: I am sorry if this seems nitpicky, and it's feedback that was given to me earlier on in this project, but being that this function is private (only used in this class), could you rename the function to _color ?

lib/pysquared/logger.py
```diff
@@ -7,9 +7,41 @@
...
+    }
+
+    _f = {"normal": "0", "bold": "1", "ulined": "4"}
+    return _h + _f[fmt] + ";3" + _c[color] + "m" + msg + _e
+
+
+LogColors = {
+    "NOTSET": "NOTSET",
+    "DEBUG": color(msg="DEBUG", color="blue"),
+    "INFO": "INFO",
```
blakejameson, line 38: I personally think that 'INFO' messages would benefit from having a color other than the default. Idk if others would agree, but although it's a unique color compared to the other log levels, I think it kinda blends in with all the other fields in the message and is somewhat more difficult to find right away

blakejameson (changes requested):
You did great with this PR Nikko! I left feedback for just a few small things, but this is an awesome job by you

lib/pysquared/logger.py
Nikko-Adrian-Pacleb: I think its good you pointed it out especially since we are trying to make the code organized. I will keep this in mind in the future.

lib/pysquared/logger.py
Nikko-Adrian-Pacleb: Do you recommend adding a color for it?

blakejameson: Yes, I think adding a color would help it. Sorry if its a hassle

lib/pysquared/logger.py
```diff
@@ -50,10 +78,18 @@ def _log(self, level: str, level_value: int, message: str, **kwargs) -> None:
...
+        json_order.update(kwargs)
+
+        json_output = json.dumps(json_order)
 
         if self._can_print_this_level(level_value):
-            print(json_output)
+            colored_json_output = json_output.replace(
+                f'"level": "{level}"', f'"level": "{LogColors[level]}"'
+            )
+            print(colored_json_output if self.colorized else json_output)
```
blakejameson, line 92: I find myself looking at this section and contemplating if it would better be refactored as something like:
```
if self._can_print_this_level(level_value):
    if self.colorized:
        json_output = json_output.replace(f'"level": "{level}"', f'"level": "{LogColors[level]}"')
    print(json_output)
```
The way the current code is, even if self.colorized is False, the code would still make an object above with color, which I think is somewhat unnecessary, as it would go on to completely disregard it and use the plain json output object. 

The proposed code idea shared here would result in the json object with color being created only when self.colorized is True.

@nateinaction @Mikefly123 Do you have any thoughts on this suggestion from me? I know that we aren't in chase of the most extreme performance but preventing unnecessary object creation doesn't hurt

nateinaction: Sounds like a good optimization, go ahead and commit it!
```suggestion
            if self.colorized:
                json_output = json_output.replace(f'"level": "{level}"', f'"level": "{LogColors[level]}"')
            print(json_output)
```

nateinaction (approved):
Looks good, thank you!

blakejameson (approved):
Great job Nikko! Congrats on your first PR

Issue #122, opened 2025-01-29
[TASK] Merge utilities.py with the Main Codebase for File System Manipulation


Mikefly123:
## Description of the Task
The hacker `LMBFAO` (an anonymous contributor) has sent to us a [utilities.py](https://github.com/proveskit/development_software/blob/main/OTA%20Support/utilities.py) file that has some working functions for manipulating the CircuitPython file system in fun ways. These functions are going to be pretty useful for logging and over the air updates and as a result we should move quickly to get them integrated with our current code. 

This task is pretty straightforward, just figure out how best to to bring these new functions into the current `pysquared` library along with any API changes that need to be made so these utilities can be used by other classes easily. 

## Acceptance Criteria
- [ ] `utilities.py` is merged with the `pysquared` library
- [ ] Some light documentation of how all the functions in `utilities.py` are used is created 
- [ ] API updated to allow functions to be used by other classes

## Technical Notes
[CircuitPython Storage Docs](https://learn.adafruit.com/circuitpython-essentials/circuitpython-storage) 
This is a bit of an open ended task. If you take it on and want more guidance ping @Mikefly123 to setup a meeting to discuss how this class is supposed to work.

PR #153, opened 2025-02-08
Merging utilities.py into repo

TaylorGilg:
## Summary
In trying to adapt utilities.py for our repo, I added comments explaining the different functions, their arguments, and returning values. I also type hinted all the functions. 

We'll have to play around with the SD to see if we can replicate the bug mentioned in line 268 and adding a mass deletion functionality is highlighted in line 326. 
Line 268: # TODO: Make file if not exist. look into why from_path file get rewritten. bug?
Line 326: # TODO: add recursive functionality to remove files

This was a very helpful resource to understand circuitpython's storage system: 
https://docs.circuitpython.org/en/latest/shared-bindings/storage/ 
I believe that enable/disable_usb_drive() can only be called in boot.py which is why I included commented code of how utilities.py was previously being called in boot.py.

This file tries to change the state of variable read_state within a parameter.json file. We can translate this however we want within the context of our current config.json structure and what permissions we wish different parts of the repo to have when it comes to changing config values. @Alandlt15 @blakejameson 

Solves Issue #122


Alandlt15:
I'm terribly sorry for the delayed response. I'm not sure if I received the notification of you @ing me. Will start looking into this ASAP. Again, my apologies.

.pre-commit-config.yaml
```diff
@@ -3,12 +3,12 @@ repos:
...
     - id: trailing-whitespace
-    # - id: end-of-file-fixer
+    - id: end-of-file-fixer
     - id: check-yaml
     - id: check-json
     - id: check-added-large-files
-    - id: mixed-line-ending
-      args: [ --fix=lf ]
+    #- id: mixed-line-ending
+    #  args: [ --fix=lf ]
```
nateinaction, line 11: Causing loads of line changes for some windows users. Commenting out for now.

.pre-commit-config.yaml
```diff
@@ -3,12 +3,12 @@ repos:
     rev: v5.0.0
     hooks:
     - id: trailing-whitespace
-    # - id: end-of-file-fixer
+    - id: end-of-file-fixer
```
nateinaction, line 6: Previously accidentally commented out, uncommenting

lib/pysquared/usb/usbfunctions.py
```diff
@@ -0,0 +1,332 @@
...
+                b = "a"  # Sets mode to append
+            with open(ff, b) as f:
+                f.tell()  # Returns current file position
+            os.chdir("/")  # Changes back to the root directory
+            return ff
+        except Exception as e:
+            print("Error creating file: " + "".join(traceback.format_exception(e)))
+            return None
+
+    def readfile(self, path: str, type: str = "string") -> str:
```
nateinaction, line 161: Expanding on `__setitem__` complaint in sonarqube. Looks like this fn can return more than just a string.

nateinaction (commented):
I feel a little iffy including this without a good understanding of how we should organize storage stuff but I haven't been able to get around to that as quickly as I would have liked. Excluding the TODO recommendations from sonarqube, if we can fix up those recommendations this should be OK to merge to keep us progressing.

nateinaction (approved):

PR #160, opened 2025-02-14
Reformat the "err" value in json output

blakejameson:
## Summary
In the Logger class, I replaced the "err" so that the value for the json output will equal traceback.format_exception(e), whereas previously "err" equaled the value of the error e. Using the format_exception() is more descriptive, as it shows the call stack.

Following Nate's advice, I also added a check for the non error severity levels (debug, info, and warning) to see if those function calls pass an argument "err" of type Exception to the kwargs. If the user does happen to pass that, the "err" json output is formatted appropriately with traceback.format_exception().

## How was this tested
- [x] Added new unit tests
- [ ] Other (Please describe)

Here is output of the code working on a v4b board, similar output to the main branch. However, it of course has the updated "err" value with the traceback.
Screenshot 2025-02-13 at 9 12 29 PM

Here is the code and output if someone uses the debug log call yet passes an Exception to a kwarg value of "err". As you can see, it formats the "err" value as a traceback
Screenshot 2025-02-13 at 8 56 31 PM
Screenshot 2025-02-13 at 8 57 47 PM

Here is the code and output if someone uses the debug log call but passes a string value to a kwarg value of "err". As you can see, it formats the "err" value as a string
Screenshot 2025-02-13 at 8 58 15 PM
Screenshot 2025-02-13 at 8 59 37 PM

### DIFF
--- a/main.py
+++ b/main.py
@@ -150,4 +150,4 @@ try:
         c.hardware["WDT"] = False
 
 except Exception as e:
-    logger.error("An exception occured within main.py", e)
+    logger.critical("An exception occured within main.py", e)
--- a/main.py
+++ b/main.py
@@ -64,7 +64,7 @@ try:
         initial_boot()
 
     except Exception as e:
-        logger.error("Error in Boot Sequence", err=e)
+        logger.error("Error in Boot Sequence", e)
 
     finally:
         pass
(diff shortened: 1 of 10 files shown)


main.py
```diff
@@ -150,4 +150,4 @@ def minimum_power_operations():
         c.hardware["WDT"] = False
 
 except Exception as e:
-    logger.error("An exception occured within main.py", err=e)
+    logger.error("An exception occured within main.py", e)
```
nateinaction, line 153:
```suggestion
    logger.critical("An exception occured within main.py", e)
```

nateinaction (approved):

nateinaction (approved):

PR #166, opened 2025-02-18
Rename Params & Methods in Big_Data to match regex

hrfarmer:
## Summary
Changed variable and method names in `Big_Data.py` to match the regex `^[_a-z][a-z0-9_]*$` that's specified in SonarQube

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)


nateinaction (approved):
Checked all var usages. LGTM

PR #167, opened 2025-02-19
Renamed uptime method in pysquared.py to get_system_uptime. 

SamuelGomezG:
Applied changes to functions.py

## How was this tested
It has not been tested. Test is needed


blakejameson (approved):

nateinaction (changes requested):
I think there are some additional spots where `self.UPTIME` is referenced in `pysquared.py`. I see at least one more on line 279. Do you see any more?

nateinaction (approved):
This is only changing the property not the member variable... feels confusing but should be OK. ✅

PR #168, opened 2025-02-19
Address (most) SonarQube issues in pysquared.py

hrfarmer:
## Summary
Went through and addressed most of the SonarQube issues that were shown, exceptions being the TODO comment and one that's addressed in #167 

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)

### DIFF
--- a/lib/pysquared/pysquared.py
+++ b/lib/pysquared/pysquared.py
@@ -12,6 +12,7 @@ import sys
 import time
 from collections import OrderedDict
 from os import chdir, mkdir, stat
+from typing import Optional
 
 import board
 import busio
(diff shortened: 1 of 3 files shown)


lib/pysquared/pysquared.py
```diff
@@ -12,6 +12,7 @@
 import time
 from collections import OrderedDict
 from os import chdir, mkdir, stat
+from typing import Optional
```
nateinaction, line 15: Unfortunately the typing library is not available in circuitpython so we'll need to wrap this in a try/except. It looks like this can be added to the existing typing import on line 38.

nateinaction (approved):

Issue #157, opened 2025-02-11
Unused functions and variables


blakejameson:
## Summary
There are some variables and functions that are defined but never utilized. 

Earlier I stopped by the Discord Meeting Room and asked @Mikefly123  if I could run the unused variables and functions by him to learn which should stay vs. which are no longer needed

## Files and their unused functions, variables
**battery_helper.py**
Functions currently not used anywhere:
•	*Currently there are no BatteryHelper function calls anywhere in the repo

**cdh.py**
Functions currently not used anywhere:
•	hotstart_handler
•	FSK

**Field.py**
Functions currently not used anywhere:
•	troubleshooting

**functions.py**
Variables not being used anywhere:
•	self.error_count (error_count was replaced by an nvm Counter object that is stored in the Logger)
•	self.detumble_enable_x
•	self.detumble_enable_y
•	self.detumble_enable_z

Functions currently not used anywhere:
•	current_check
•	send_packets
•	listen_joke
•	get_battery_data
•	OTA

**packet_sender.py**
Functions currently not used anywhere:
•	handle_retransmit_request
•	fast_send_data

**pysquared.py**
Functions currently not used anywhere:
•	reset_vbus
•	powermode
•	print_file
•	read_file
•	new_file

Mikefly123:
Okay, here's a start at outlining what some of these unused functions do!

**battery_helper.py**
This was a helper class for interacting with the battery board. Note how there are a bunch of functions in there just supporting board to board UART communications. We could just nuke the whole thing, or leave it in and just leave it for now and refactor it for the upcoming V2 updates.

**cdh.py**
- `hotstart_handler` this is a function that indeed is currently unused, but is intended as a helper function to ease the satellite back into normal operations after a hibernation period. I would leave it for now until the `rtc`and `radio` library updates are done and we reimplement genuine hibernation functionality.
- `FSK` I believe this is in to assist in switching between `lora` and `fsk` modes on the satellite. I think this should be refactored to align with Nate's new radio modulation manager.

Mikefly123:
Some more explanation of the unused functions: 

**Field.py**
- troubleshooting: I think this function can just be removed, I think its an artifact of when we first setup that class. Eventually we will refactor this class into some kind of packet factory.

**functions.py**
- `error_count`: since the way we track this has been changed, I think we can remove this function from `functions.py`
[...]

PR #169, opened 2025-02-19
Deleted battery_helper.py

asiemsen:
## Summary
Sonarqube identified [an issue](https://sonarcloud.io/project/issues?open=AZSf6hbnjG4pb3pY2AeA&id=proveskit_circuitpy_flight_software) in `battery_helper.py`. Based on Michael's comment in #157 he mentioned we could nuke it. Choosing to nuke it instead of fixing it.

## How was this tested
no parts of battery_helper class is ever used.


blakejameson (changes requested):
Great work Aaron
I noticed that `battery_helper` is imported from in functions.py and that looks to be the only spot it is mentioned 

Could you delete the import line? You should be set after that

asiemsen:
Thanks for the catch!
Looking into the file a little further I'm seeing that 'battery_helper' is actually used to instantiate 'self.battery' which is used in 'get_battery_data' to return the power metrics of the battery. This function is currently not being called anywhere as far as I can tell. Would you recommend that I get rid of 'self.battery' and the 'get_battery_data' function in addtion to the import?

blakejameson:
@asiemsen Look at you with the great catch
I think it would be a great idea to get rid of 'self.battery' and the 'get_battery_data' function 
Awesome job!

nateinaction (approved):

blakejameson (approved):

PR #175, opened 2025-02-21
Identify currently running version

### LINKED ISSUE #171: [FEATURE] Identify the Current Version of Software in the Code
## Story
* **As a** flight controller board user.
* **I want** to know which version of software is currently loaded onto my board.
* **So that** I can quickly identify if my current configuration has any known issues or special features.

## Acceptance Criteria

- [ ] The flight controller board displays the current software version number when powered on in the serial output
- [ ] The displayed version number matches the actual software installed on the board
- [ ] The version number is stored as a datatype on the satellite that can be evaluated in runtime (i.e. we can use code to check if the version loaded is the version we expected)

## Technical Details
As we work on trying to get Over The Air (OTA) updates implemented into our code, we need a way to check what version of the code is currently loaded. This will allow us to effectively swap between different versions of software in the future when we have an onboard updating system.

At the moment we are [doing weekly releases](https://github.com/proveskit/circuitpy_flight_software/releases) of the software that is on `main` so there are unique version numbers for each week. I think it would be a good place to start by just looking to implement those version numbers into the software somehow.

hrfarmer:
Closes #171 

## Summary
This adds a VERSION variable to the Makefile, which runs a git command to check if the current commit points to a tag (since releases are tagged), and set the variable as such. The build step then writes the variable to version.py, which can then be pulled from.

I needed to put the modified version file inside of the rsync call, because otherwise on non-Windows systems, the modified file wouldn't be copied to the board since rsync is directly called instead of copying everything inside of the artifact folder. 

I added it to the Satellite class so that it can be accessed from there easily, let me know if I should store it somewhere else instead.

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)

image
(shows an empty string since the current commit doesn't point to a tag)


hrfarmer:
ci test is failing because of adding the file directly to the rsync call, so that probably isn't a good direction. 

I noticed that in the install command on the makefile, it directly copies the artifact folder (which would work) for Windows, but then calls rsync directly for other platforms, which means the modified file won't get copied to the board (but will be in the build folder). Is there a specific reason that's done, or could that if check be removed and a direct copy is done on all platforms?

Alternatively, I could instead just move the logic into the rsync call and use a temp file, so the original file isn't overwritten by the new one

Lex-ari (approved):
Looks good! Just need to fix rsync

Lex-ari (approved):
lgtm

Mikefly123 (approved):
LGTM! 

I think to accompany this we should consider next week reformatting the repo to have the `main` branch only contain "stable" release tags and all of the development goes into a `devel` branch.

PR #177, opened 2025-02-21
Add version field to repl.py

hrfarmer:
## Summary
Oversight from the version pr, forgot to add the version to the Satellite class in repl.py, so it was erroring

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)


blakejameson (approved):
Thanks so much for updating! Tested in REPL and it worked

PR #178, opened 2025-02-21
Eliminated commented out code.

SamuelGomezG:
## Summary
Eliminated the unused code in `debugcolor.py` from line 1 to line 4.
https://github.com/proveskit/circuitpy_flight_software/blob/124907cbf7fec6c939021ce117b86c07ca32ec40/lib/pysquared/debugcolor.py#L1

### DIFF
--- a/lib/pysquared/debugcolor.py
+++ b/lib/pysquared/debugcolor.py
@@ -1,8 +1,3 @@
-"""
-from debugcolor import co
-print(co(msg='hello!',color='red'))
-"""
-
 _h = "\033["
 _e = "\033[0;39;49m"
 


blakejameson:
It looks like the purpose of the commented out code is demonstrating how this class would be used. Going to pull in @Mikefly123 to see what he thinks, but my gut feeling is he will think it should stay as is

Mikefly123:
@blakejameson is correct here that the comment string in question is an example of how the code is intended to be used. I think it's appropriate to embed one, since the underlying class is not very human readable. 

I'm going to block this merge for now because Nikko is refactoring the use of Debug Color in #148 anyways.

Issue #170, opened 2025-02-19
[FEATURE] Utilize RP2040's built-in RTC to generate timestamp in Logs


blakejameson:
## Story
* **As a** `circuitpy_flight_software` user
* **I want** to be able to see an accurate current time and date in log output
* **So that** I am able to know at the precise time events occur

Currently, the logger outputs a time and date that is based on time.localtime(). 
Having the logger use the RTC instead would allow for accurate time and date output.
* EDIT: time.localtime() and time.time() is[ backed by the RP2040's RTC ](https://docs.circuitpython.org/en/latest/shared-bindings/rtc/#module-rtc)

## Acceptance Criteria

- [x] Uses the RP2040's RTC to retrieve the current time and date
- [x] Logs output with a timestamp based on RTC 

## Technical Details

The Logger class will be refactored to utilize the RV3028 module

blakejameson:
@Mikefly123 In the process of working on this issue, I realized that the RTC object is created in the Satellite object in pysquared.py

Of course the Logger would need to access the RTC object, but being that the Satellite has a parameter it takes in of Logger, the Satellite object can't then be passed into the Logger object

A potential solution is initializing the RTC and I2C1, as the RTC depends on it, in main.py. Then passing the RTC to the Logger, and the I2C1 and RTC to Satellite

I was curious on what your thoughts were with this approach or if you recommend another way to go about it

Mikefly123:
Hey @blakejameson! I think a good way around this issue is to use the built in RTC that the RP2040 has as the means of time stamping the logs: 

https://docs.circuitpython.org/en/latest/shared-bindings/rtc/

This RTC would not have an way of sustaining itself between resets (our external RTC theoretically would, but the library support for that is not done and tested yet), so you would need to resync it after every reset and probably also periodically check if the time is drifting (internal microcontroller clocks are notorious for drifting over time).

blakejameson:
@Mikefly123 Thanks again for sending me this link! Using the built in RTC has been a huge help already

Currently in the early stages but can now enter the REPL to set the time and date and then control-D to soft reset and I see the proper time and date values for log output timestamps.

As you can see at the bottom of the output, the check_reboot() function results in an uptime of a huge number (162,247,351 seconds in this case) that far exceeds the 3600 second reboot time set in the config. This of course causes a hard reset

It looks like I will get to refactoring the BOOTTIME and other related logic to prevent this. 

[...]

Issue #139, opened 2025-02-04
[FEATURE] Set RTC Time Using a `make` Command


Mikefly123:
## Story
* **As a** satellite developer.
* **I want** for my board to know the correct time.
* **So that** logs and other time sensitive information is correctly timestamped. 

Currently our log files are all being timestamped with a nebulous mystery time that the microcontroller is keeping. Wouldn't it be cool for this to be the actual real world time? 

## Acceptance Criteria

- [ ] Establish a method for a `make sync-time` command to automatically open a serial connection with the flight controller board and enter the REPL
- [ ] Call the `rtc` time and date setter functions to set the RTC time to the current UTC time. 

## Technical Details
On flight controller board V4b and above there is an RTC (Real Time Clock) IC that is onboard for keeping time! There are still some features not yet introduced into the drivers (like sustaining the time after being powered down and wake up alarms), but within a single power up the time does seem to be kept pretty correctly!

nateinaction:
Right now `make install` only needs to know about the filesystem mount point. We would need to make the command also know about the tty/serial port. I suspect the behavior will also be different between the posix systems and windows.

Maybe we could start with a separate `make sync-time` command to feel out an implementation?

Mikefly123:
Yeah, I could see an independent command making sense! I'll update the original ticket to reflect that.

blakejameson:
I tried to utilize the RTC methods but experienced the bug discussed in https://github.com/proveskit/PROVES_CircuitPython_RV3028/issues/4

<img width="539" alt="Image" src="https://github.com/user-attachments/assets/27f79f39-27e4-4faf-be41-bc027f7dbc43" />

blakejameson:
Hey @Mikefly123, Quick question:

In https://github.com/proveskit/circuitpy_flight_software/issues/170#issuecomment-2686467817, you mention "check the rtc current time and if it is more than a few seconds off, it should sync and then notify the user that the time was off by x amount and needed to be synced."

Should the acceptance criteria of this issue be updated to include that? Or should this issue only deal with setting the time when the `make sync-time` command is called and a subsequent issue would deal with detecting the difference of times and only updating if the difference in times is large enough?

blakejameson:
@Mikefly123 Sorry to @ you again but you may have missed this previous question I asked. Let me know if you get the chance

Mikefly123:
[...]

PR #179, opened 2025-03-04
170 utilize rtc to generate timestamp in logs

blakejameson:
## Summary

This PR dealt with allowing functionality to update the RP2040's RTC. Updating the RP2040's RTC allows for much better insight regarding log messages. 

To use the new function to set update the RP2040's RTC, enter the REPL. Once in the REPL, enter:
`from lib.pysquared.pysquared import set_rp2040_rtc_time` then
call the `set_rp2040_rtc_time` function.

**An example of using the function:** Setting the updated time and date on the board to Monday, March 3, 2025 at 6:52 pm would look like:
`set_rp2040_rtc_time(2025, 3, 3, 18, 52, 0, 1)`

After calling the function, control+D as normal to reboot the program. The time and date should continue to update as long as the board stays powered. Once unplugged, the board will default to January 1, 2020 upon booting.

As I mention in https://github.com/proveskit/circuitpy_flight_software/issues/170#issuecomment-2692963026, the RP2040's RTC can run a bit fast at times. 
A next issue, [make sync-time](https://github.com/proveskit/circuitpy_flight_software/issues/139), will be for detecting when there is a big enough disparity of the current time in real life and the time on the board. That would prompt a command to update the time on the board programmatically, as opposed to this current manual approach

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)
Screenshot 2025-03-03 at 6 44 19 PM


nateinaction:
Added tests in #184. What do you think?

blakejameson:
> Added tests in #184. What do you think?

@nateinaction Just approved 184. Amazing job

nateinaction (approved):

nateinaction:
Nice work @blakejameson! Thanks for figuring this out!

Issue #180, opened 2025-03-04
Outdated function call of supervisor.set_next_stack_limit(4096 + 4096) in boot.py


blakejameson:
## Summary
I learned that the `supervisor.set_next_stack_limit(4096 + 4096)` function call in boot.py isn't a usable function in versions of CircuitPython >= 8.0.0

## Context
Yesterday, I spent some time exploring the writing of files onto filesystem and that prompted me to make slight changes in boot.py. In boot.py, I found myself curious as to what the `supervisor.set_next_stack_limit()` function call did. In searching the function call online, I didn't see it in the [supervisor module](https://docs.circuitpython.org/en/latest/shared-bindings/supervisor/index.html#supervisor.Runtime.next_stack_limit). I then found an [Adafruit 8.0.0 release post](https://blog.adafruit.com/2023/02/06/circuitpython-8-0-0-released/) that shared a property called `supervisor.runtime.next_stack_limit` replaced `supervisor.set_next_stack_limit()` in CP 8.0.0. However, I didn't see `supervisor.runtime.next_stack_limit` in the current version of the [supervisor module](https://docs.circuitpython.org/en/latest/shared-bindings/supervisor/index.html#supervisor.Runtime.next_stack_limit), so I am not sure if that is usable anymore as well.

Lastly, I realized that there's a file I have always overlooked in boot_out.txt and the output of the file showed me the following message
<img width="930" alt="Image" src="https://github.com/user-attachments/assets/8d02804c-80c6-471c-9f3e-26949f53d0a3" />

Anyways, I think it's fair to say that the `supervisor.set_next_stack_limit(4096 + 4096)` isn't altering anything currently. Just wanted to bring attention to this. Am not sure what the next steps would be, if any

Mikefly123:
Interesting find, thanks Blake! I couldn't find a reference to this function in the api docs, but I distinctly remember when we were on CircuitPython 7.X.X it allowed you to define how big the stack was allowed to get in the Python VM. This was useful because limiting the size of the stack could say oodles of memory on the heap for scary dynamic allocations like loading an image into RAM. 

This was mainly driven by needing to protect a large part of the heap for storing and forwarding images taken by a PROVES Kit Camera. Annoyingly, it doesn't seem like there is any great documentation on this stuff anymore (if it even existed at all in the first place). It might be best at this point to just get rid of this functional call and not think too much about memory manipulation at this stage

PR #183, opened 2025-03-04
Removed supervisor.set_next_stack_limit() from boot.py

blakejameson:
## Summary
I removed the `set_next_stack_limit()` call from boot.py

As mentioned in https://github.com/proveskit/circuitpy_flight_software/issues/180#issue-2895261546, the `set_next_stack_limit()` call was removed in CP 8.0.0. The following boot.txt output showed when running code on the board with that line in boot.py:
Screenshot 2025-03-04 at 1 33 44 PM

Then when using the property that replaced it in CP 8.0.0, `supervisor.runtime.next_stack_limit`, an error occurred as well. It looks like that property is no longer supported either in CP 9
Screenshot 2025-03-04 at 2 47 41 PM

In following Michael's advice in https://github.com/proveskit/circuitpy_flight_software/issues/180#issuecomment-2698768875, "It might be best at this point to just get rid of this functional call and not think too much about memory manipulation at this stage", I am removing the function call in this PR.

Just to be safe, I ran the code on the board with this change and it runs as normal:
Screenshot 2025-03-04 at 3 00 33 PM

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -1,3 +1,0 @@
-import supervisor
-
-supervisor.set_next_stack_limit(4096 + 4096)


Mikefly123 (approved):
LGTM!

PR #184, opened 2025-03-06
170-2 Split from pysquared, add tests, add test debug to launch config

nateinaction:
## Summary
This PR modifies #179 to split functionality out of pysquared. This allows us to more easily add tests (although adding tests for the RTC was a little challenging). I'm not sure if organization in the `rtc` module is overkill but feels OK for now. Added a debug config to help diagnose testing issues.

**Testing Challenges:**
I had some issues deciding on which components in this test to mock and how to mock them. I started with trying to test the `rtc_common.py` file which I felt was simpler since we only needed to know if the fns had been called not concerning ourselves with setting any values.

I first attempted to use MagicMock's call counter but had difficulty patching the `rtc` module since it was a singleton. Next I tried to create my own RTC mock and found that mocking out the singleton was a little challenging. Success only came from dropping into the debugger to walk through the tests and ensure we weren't using the same memory address between tests among many other issues I encountered.

Even though these tests were challenging to create the result is appears to be simple and _should_ require little maintenance 🤞.

## How was this tested
- [x] Added new unit tests
- [ ] Other (Please describe)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,18 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Unit test",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "${workspaceFolder}/.venv/bin/pytest",
+            "args": [
+                "tests/unit",
+                "--maxfail=1",
+                "--disable-warnings"
+            ],
+            "console": "integratedTerminal",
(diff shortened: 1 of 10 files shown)


blakejameson (approved):
You did an exceptional job here. Being able to implement the mocks and tests is so huge. Kudos Nate, kudos

Issue #108, opened 2025-01-24
Add the watchdog timer as our first asyncio component


nateinaction:
## Description

**As a** flight software engineer
**I want** to have an established pattern for running multiple tasks at a time
**So that** we can adopt the pattern for other components moving forward

The watchdog is sufficiently isolated component that it might be a good candidate for an initial and simple asyncio implementation.

## Acceptance Criteria

- `watchdog_pet()` is only called in one location (possibly a while loop in `main.py`)
- remove additional calls to `watchdog_pet()` in `main.py` and `safe_sleep()`
- `watchdog_pet()` is running asynchronously in the background

## Technical Details
- @Mikefly123 does a good job explaining what the hardware watchdog is and how it works on proveskit/fprime-proves#27
- Not the same asyncio library but this video shows what it might be like to write a simple program using asyncio for micropython: https://www.youtube.com/watch?v=5VLvmA__2v0
- I could be wrong but suspect successful completion of this task will involve importing asyncio from lib in `main.py` and `pysquared.py` and modifying `main.py` and the following lines [watchdog_pet](https://github.com/proveskit/circuitpy_flight_software/blob/main/lib/pysquared/pysquared.py#L661-L664)
- Calls to remove in [safe_sleep](https://github.com/proveskit/circuitpy_flight_software/blob/main/lib/pysquared/functions.py#L69-L84)

Mikefly123:
It should be relatively straight forward to implement! I think main thing to figure out will be how to reconcile having asyncio `awaits` in multiple difference locations in the code. We originally only had it implemented in `main` and it worked fine, but when we transitioned to the new `adafruit_RFM` radio library the implementation broke because the new radio library also has `await` calls that seemed to cause a race condition with `main`.

PR #189, opened 2025-03-07
[#108] Update watchdog_pet()

yudataguy:
## Summary

The `watchdog_pet` function has been updated, it's now async running in background in the main loop. Hope this satisfy #108 

## How was this tested

Code is not tested yet due to watchdog complexity, will hand off to @nateinaction for further testing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ ifeq ($(OS),Windows_NT)
 	rm -rf $(BOARD_MOUNT_POINT)
 	cp -r artifacts/proves/* $(BOARD_MOUNT_POINT)
 else
+	@rm $(BOARD_MOUNT_POINT)/code.py > /dev/null 2>&1 || true
 	$(call rsync_to_dest,$(BOARD_MOUNT_POINT))
 endif
 


main.py
```diff
@@ -66,78 +68,41 @@
...
-        c.watchdog_pet()
-
-        sleep_helper.short_hibernate()
-
     ######################### MAIN LOOP ##############################
     try:
         while True:
             # L0 automatic tasks no matter the battery level
-            c.check_reboot()
+            c.check_reboot()  # This now includes watchdog petting
```
Mikefly123, line 105: This approach doesn't actually work for checking the reboot (and therefore petting the watchdog) in the background because the lower down `main()`, `critical_power_operations()`, and `minimum_power_operations()` are functions that will hold the thread until all of the things they are trying to do are completed. 

In the future we might have to change the entire `main.py` setup so everything is once again task based rather than hard coded sequences, but as a stopgap for now maybe we can `import asyncio` in here and try to allow the `main()` function to release the thread by adding `await` statements in front of every call in the sequence? My preferred AI chatbot recommends something like this:

```py

import asyncio

# Define your sequence of functions as coroutines
async def func1():
    print("Running func1")
    await asyncio.sleep(2)  # Simulate work

async def func2():
    print("Running func2")
    await asyncio.sleep(2)  # Simulate work

async def func3():
    print("Running func3")
    await asyncio.sleep(2)  # Simulate work

# Define the watchdog function as a coroutine
async def watchdog():
    print("Watchdog function called")
    await asyncio.sleep(1)  # Simulate its work

# Function to run the sequence with a watchdog in between
async def run_sequence_with_watchdog(functions):
    for func in functions:
        await func()  # Run the main function
        await watchdog()  # Run watchdog
        await asyncio.sleep(1)  # Release control for 1 second

# List of functions to execute in sequence
functions = [func1, func2, func3]

# Run the sequence
asyncio.run(run_sequence_with_watchdog(functions))
```

Mikefly123 (changes requested):
I verified that all of the new functions in `pysquared.py` work, so good job getting those together! The setup of the watchdog as a background task doesn't work in `main.py` right now though, so we need to consider either making a new ticket to return everything in `main.py` from being a static sequence to a task based execution system or making those changes in this PR directly.

main.py
Mikefly123: I think you have a V4c board. If so, whenever the watchdog is pet, a little red LED in the bottom right side of the board should flash. You can see the video linked in [this pull request](https://github.com/proveskit/fprime-proves/pull/45) for an example of what this looks like.

Mikefly123: @yudataguy yes that is the correct way to test it! Make sure you are overwriting the old file when doing this. 

I'm not quite sure what the current state of your dev board is, but I went ahead and merged our change from #190 into your branch. If you use `make install` to do a clean install from your branch it should now make it so `main.py` runs by default. It is possible you were not seeing anything happen because of the bug mentioned where sometimes `code.py` would just do nothing.

SamuelGomezG:
I wanna try and see how it is that we could transform routines to tasks so that we can really take advantage of asyncio.

main.py
Mikefly123: Yeah @yudataguy it does indeed look like the watchdog is being pet! Can you verify that all of the other functions happening in the board's main loop are also working as intended? It should tick through a few things before going into a cycle of trying to transmit and safe sleep.

Mikefly123 (approved):
Hey @yudataguy that's actually a new bug we hadn't noticed before! I don't think it has anything to do with your code. I opened a fresh ticket to track it in #204. 

I concur things look good from your screenshots! I think we will soon try to refactor this into an upcoming `state_of_health_manager`, but I will go ahead and approve this merge so we can get this functionality onboard in the meantime!

PR #190, opened 2025-03-08
Fixed Missing `code.py`

Mikefly123:
## Summary
When using code from the `main` branch, nothing happens when you do a `make install` on a clean board.

I'm not sure when it happened, but somehow `code.py` got nuked from the `main` branch. This is an issue because on clean installs of the CircuitPython Firmware a `code.py`with only a `print("Hello World") statement is generated automagically.

Because `make install` doesn't overwrite this `code.py` with anything on startup the board will only execute the contents of `code.py` and exit to the REPL without executing any of the code in `main.py`. If this is the intended behavior we need to document it as such, but I suspect this is not intended behavior because if you attempt to `import main` from the REPL it will throw a critical `WDT_WDI in use` error:

Screenshot 2025-03-07 at 4 12 57 PM

With this patch we go back to normal functionality.

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)

## Adafruit Context
https://learn.adafruit.com/welcome-to-circuitpython/creating-and-editing-code

Screenshot 2025-03-07 at 4 18 01 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ ifeq ($(OS),Windows_NT)
 	rm -rf $(BOARD_MOUNT_POINT)
 	cp -r artifacts/proves/* $(BOARD_MOUNT_POINT)
 else
+	@rm $(BOARD_MOUNT_POINT)/code.py > /dev/null 2>&1 || true
 	$(call rsync_to_dest,$(BOARD_MOUNT_POINT))
 endif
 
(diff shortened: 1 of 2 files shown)


Alandlt15 (approved):
LGTM! Really great catch there. Although I don't have a clean board, I made sure the code was running fine.

nateinaction (changes requested):
There was a problem with having a code.py file in the root. Let me see if I can recreate it and give context for its deletion.

nateinaction (approved):

PR #192, opened 2025-03-08
rm code.py in make install

nateinaction:
## Summary
Unfortunately, it looks like Python has a built-in module called `code` which causes the IDE to freak out a bit.
Screenshot 2025-03-08 at 10 46 34

It also seems to prevent our code coverage analysis from running.
Screenshot 2025-03-08 at 15 19 46
https://github.com/proveskit/circuitpy_flight_software/actions/runs/13731659762/job/38429209618

One option is to delete `code.py` while running `make install`. Not perfect but might do in a pinch. What do you think @Mikefly123 

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ ifeq ($(OS),Windows_NT)
 	rm -rf $(BOARD_MOUNT_POINT)
 	cp -r artifacts/proves/* $(BOARD_MOUNT_POINT)
 else
+	@rm $(BOARD_MOUNT_POINT)/code.py > /dev/null 2>&1 || true
 	$(call rsync_to_dest,$(BOARD_MOUNT_POINT))
 endif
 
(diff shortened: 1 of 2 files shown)


Mikefly123:
A bit of a hacky fix but this works for me! I'll go ahead and use this for #190. 

In the future when we start compiling fresh CircuitPy firmware again maybe we can see if there's an option to disable the default code.py generation.

nateinaction:
@Mikefly123 I did try to looking into that but I wasn't able to find a way to do it today. I also tried setting the default first code file looked at by circuitpython to `main.py` in `boot.py` via [`supervisor.set_next_code_file()`](https://docs.circuitpython.org/en/latest/shared-bindings/supervisor/index.html#supervisor.set_next_code_file) but it didn't work. I think you're right the answer is probably somewhere in the firmware build...

PR #193, opened 2025-03-12
Set CPU clock speed with microcontroller pkg

nateinaction:
## Summary
We found that the RP2040 does not need a custom `machine` firmware package to change the clock speed. We can use CircuitPython's built-in [`microcontroller.Processor.frequency = some value`](https://docs.circuitpython.org/en/latest/shared-bindings/microcontroller/index.html#microcontroller.Processor.frequency) API. As discussed in https://github.com/proveskit/flight_controller_board/issues/29

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)

### DIFF
--- a/lib/pysquared/pysquared.py
+++ b/lib/pysquared/pysquared.py
@@ -16,7 +16,6 @@ from os import chdir, mkdir, stat
 import board
 import busio
 import digitalio
-import machine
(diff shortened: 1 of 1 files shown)


blakejameson (approved):
Great work Nate

Issue #195, opened 2025-03-13
[BUG] "Free Memory Stat at beginning of all_face_data function" call shows low memory


blakejameson:
## Summary
I noticed that after the "Free Memory Stat at beginning of all_face_data function" showed a low number running throughout the program. It confused me why the number showed as low as it did and why it was lower than the "Free Memory Stat after importing Big_data library" message

A few screenshots of running the main branch on a v4b board:
<img width="988" alt="Image" src="https://github.com/user-attachments/assets/21b60a4b-3f14-4116-b042-9f681b2af822" />
<img width="964" alt="Image" src="https://github.com/user-attachments/assets/04b5cdd2-940d-45cd-a1af-0cbad65acedb" />

## Steps to reproduce
Run the code normally on a board

PR #196, opened 2025-03-13
moved gc .collect() in order for the free memory stat to be more accurate

blakejameson:
## Summary
Previously, the "Free Memory Stat at beginning of all_face_data function" call showed bytes_free to be low. Not only was it low but it was also lower than the subsequent call of bytes_free for "Memory Stat after importing Big_data library". See https://github.com/proveskit/circuitpy_flight_software/issues/195#issue-2917934904 for the previous behavior 

I realized that the gc.collect() wasn't called until after the "Free Memory Stat at beginning of all_face_data function" line was logged, so the "Free Memory Stat at beginning of all_face_data function" line wasn't accurate, as the existing garbage hadn't been collected.

With the change I made, now the "Free Memory Stat at beginning of all_face_data function" line is accurate, as it truly represents how much memory is currently available, as garbage had been collected a line above.

The new output of "Free Memory Stat at beginning of all_face_data function" making the change I did:
Screenshot 2025-03-13 at 12 48 12 PM
Screenshot 2025-03-13 at 12 42 26 PM
Screenshot 2025-03-13 at 12 39 06 PM
Screenshot 2025-03-13 at 12 35 47 PM

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)

### DIFF
--- a/lib/pysquared/functions.py
+++ b/lib/pysquared/functions.py
@@ -260,11 +260,11 @@ class functions:
 
     def all_face_data(self) -> list:
         # self.cubesat.all_faces_on()
+        gc.collect()
         self.logger.debug(
             "Free Memory Stat at beginning of all_face_data function",
             bytes_free=gc.mem_free(),
         )
-        gc.collect()
 
         try:
             import lib.pysquared.Big_Data as Big_Data


Alandlt15 (approved):
Looks good and works good on board! Great job Blake!

PR #200, opened 2025-03-18
Getting rid of the Field class

SamuelGomezG:
1. The functionality of the `Field` class was moved to the `radio manager` class in the method `beacon_radio_message`.
2. Every instance of the `Field` class in `functions.py` was deleted along with imports, deletions, and uses of the garbage collector. Replaced every instance for a call to `beacon_radio_message` method from `radio manager` class.
3. Created a test for `beacon_radio_message` method in `test_manager.py`. Test was successful.

## How was this tested
- [X] Added new unit tests


lib/pysquared/pysquared.py
nateinaction, line 1: Looks like these changes relate to some of your asyncio work and might have been committed on accident.

main.py
nateinaction, line 1: Same with the changes in this file.

nateinaction (changes requested):
Looks like a nice simplification! Check to make sure you intended to commit `pysquared.py` and `main.py`.

Issue #27, opened 2024-08-02
Improve development speed by linting for circuit python spec


nateinaction:
Recently tried running #26 on the satellite and received the following error `no module named 'dataclasses'`. We should try to lint to ensure all code adhears to the ~micropython~ circuit python standard so devs receive quick feedback to know that their code will work at least run when uploaded to the satellite. I suspect this would have saved me about an hour.

nateinaction:
Research notes:
Started off with a [blog article](https://lemariva.com/blog/2019/08/micropython-vsc-ide-intellisense) describing a vscode micropython linting setup which lead me to [micropy-cli](https://github.com/BradenM/micropy-cli). micropy-cli aims to offer linting, intellisence, autocompletion, and dependency management for free as long as you can specify the board you're using. I then started trying to understand how to specify our custom board definition and with Michael's help was pointed to the [CircuitPython build data](https://github.com/proveskit/flight_controller_board/tree/main/Firmware/pysquared_rp2040_flight_controller_v1a) for the Proveskit board. Finally I found the [CircuitPython Stubs](https://pypi.org/project/circuitpython-stubs/) package and a pretty [recent issue](https://github.com/adafruit/circuitpython/issues/8756) describing how an imperfect but serviceable way to set up a custom pinout.

nateinaction:
Ultimately what I have found is that you can install circuit python stubs with the [`circuitpython_setboard`](https://learn.adafruit.com/welcome-to-circuitpython/pycharm-and-circuitpython#device-specific-board-stubs-3168661) command. However your board needs to be included in a circuitpython release. The circuitypython team has a guide on how to add your board [here](https://learn.adafruit.com/how-to-add-a-new-board-to-circuitpython/overview)

nateinaction:
Contributing the v4 board definition to the circuitpython repo is in-flight but waiting on our USB VID:PID application.
- Application: https://github.com/pidcodes/pidcodes.github.com/pull/1007
- Board definition introduction branch: https://github.com/adafruit/circuitpython/compare/main...proveskit:circuitpython:proveskit-rp2040-v4?expand=1

PR #201, opened 2025-03-19
Use circuitpython 9.2.5 for v4 board stubs

nateinaction:
## Summary
This PR updates the board import in `main.py` to use the upstream board stub in the IDE. We can now see exactly which pins are available to use. Partially completes #27

I did not update all locations where `board` is imported since I believe we should likely limit usages of `board` to `main` or other files that do not live in `pysquared`. Especially as we add the v5 board, I think `pysquared` will become the board independent library for satellite development. 

Screenshot 2025-03-19 at 18 09 49

Screenshot 2025-03-19 at 18 09 38

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)

Change only impacts IDE, will not run on board.

### DIFF
--- a/main.py
+++ b/main.py
@@ -11,9 +11,13 @@ Published: Nov 19, 2024
 import gc
 import time
 
-import board
 import digitalio
 import microcontroller
+
+try:
+    from board_definitions import proveskit_rp2040_v4 as board
+except ImportError:
+    import board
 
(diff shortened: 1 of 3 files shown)


Mikefly123 (approved):
LGTM!

I also concur on trying to limit where we use the `board` module.

PR #202, opened 2025-03-20
Getting Rid of Field Class

SamuelGomezG:
1. The functionality of the `Field` class was moved to the `radio manager` class in the method `beacon_radio_message`.
2. Every instance of the `Field` class in `functions.py` was deleted along with imports, deletions, and uses of the garbage collector. Replaced every instance for a call to `beacon_radio_message` method from `radio manager` class.
3. Created a test for `beacon_radio_message` method in `test_manager.py`. Test was successful.

## How was this tested
- [X] Added new unit tests


lib/pysquared/functions.py
```diff
@@ -126,13 +118,8 @@ def beacon(self) -> None:
...
             )
 
-        self.field: Field.Field = Field.Field(
-            self.cubesat, self.logger, self.radio_manager
-        )
-        self.field.Beacon(lora_beacon)
-        del self.field
-        del Field
-        gc.collect()
+        # self.field.Beacon(lora_beacon)
```
nateinaction, line 121:
```suggestion
```

Looks like we can remove this too.

lib/pysquared/functions.py
```diff
@@ -80,20 +80,13 @@ def send(self, msg: Union[str, bytearray]) -> None:
...
             msg (String,Byte Array): Pass the String or Byte Array to be sent.
         """
-        import lib.pysquared.Field as Field
-
-        self.field: Field.Field = Field.Field(
-            self.cubesat, self.logger, self.radio_manager
-        )
         message: str = f"{self.callsign} " + str(msg) + f" {self.callsign}"
-        self.field.Beacon(message)
+        # self.field.Beacon(message)
```
nateinaction, line 84:
```suggestion
```

Looks like we can remove this comment.

nateinaction (commented):
Looks pretty good. Just removing two commented lines.

Issue #205, opened 2025-03-24
[BUG] Stuck petting watchdog


blakejameson:
## Summary
I noticed that after 3 minutes and 42 seconds into running, the main loop is blocked by the watchdog pet and the "petting watchdog" output is the only thing that is output 

The following output occurred for 6 minutes more

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/659b8f03-82b9-476e-9d04-a4dbbb3e7143" />

## Steps to reproduce
Run the code from the main branch

Mikefly123:
Hey @blakejameson, can you verify the config of your board when you ran into this issue? iirc you were testing on a V4b, I was unable to reproduce this bug on a V4c using `CircuitPython 9.1.4`.

See the logs from my test run of the `main` branch here:

```json
Auto-reload is on. Simply save files over USB to run them or enter REPL to disable.
code.py output:
{"time": "2020-01-01 00:02:10", "level": "INFO", "msg": "Booting", "published_date": "November 19, 2024", "software_version": ""}
{"time": "2020-01-01 00:02:10", "level": "INFO", "msg": "Code Starting in 5 seconds"}
{"time": "2020-01-01 00:02:11", "level": "INFO", "msg": "Code Starting in 4 seconds"}
{"time": "2020-01-01 00:02:12", "level": "INFO", "msg": "Code Starting in 3 seconds"}
{"time": "2020-01-01 00:02:13", "level": "INFO", "msg": "Code Starting in 2 seconds"}
{"time": "2020-01-01 00:02:14", "level": "INFO", "msg": "Code Starting in 1 seconds"}
{"time": "2020-01-01 00:02:15", "level": "DEBUG", "msg": "Initializing Config"}
{"time": "2020-01-01 00:02:15", "level": "DEBUG", "msg": "Booting up!", "boot_time": "1577836935s"}
{"time": "2020-01-01 00:02:15", "level": "DEBUG", "msg": "Initializing hardware component", "hardware_key": "I2C0"}
{"time": "2020-01-01 00:02:15", "level": "DEBUG", "msg": "Hardware component not initialized", "cubesat": "Orpheus", "hardware_key": "I2C0"}
{"time": "2020-01-01 00:02:15", "level": "DEBUG", "msg": "Initializing hardware component", "hardware_key": "SPI0"}
{"time": "2020-01-01 00:02:15", "level": "DEBUG", "msg": "Initializing hardware component", "hardware_key": "I2C1"}
[...]

PR #209, opened 2025-03-25
Revert "[#108] Update watchdog_pet()"

blakejameson:
Reverts proveskit/circuitpy_flight_software#189

Currently, when running the code on the main branch, it eventually gets to a point where the only output is "petting watchdog". The other code in the main loop isn't showing that it's running whenever this occurs

Screenshot 2025-03-24 at 12 17 14 PM

### DIFF
--- a/lib/pysquared/pysquared.py
+++ b/lib/pysquared/pysquared.py
@@ -8,7 +8,6 @@ Library Repo:
 """
 
 # Common CircuitPython Libs
-import asyncio  # Add asyncio import
 import sys
 import time
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


Mikefly123:
Hey @blakejameson I'm going to spend a bit of time today trying to see if we can fix the bug you described in #205. I'm not sure yet, but my best guess is that when we do go to `safe_sleep` it somehow is nuking the asyncio instance in `main`.

```py
        while duration >= 15 and iterations < 12:
            time_alarm: circuitpython_typing.Alarm = alarm.time.TimeAlarm(
                monotonic_time=time.monotonic() + 15
            )

            alarm.light_sleep_until_alarms(time_alarm)
            duration -= 15
            iterations += 1

            self.cubesat.watchdog_pet()
 ```
 Will report back after a bit of testing.

Mikefly123 (commented):
See my comment in #205. I was unable to reproduce the bug that leads to this PR. Might just be a config issue or maybe an incompatibility between different versions of CircuitPython?

Mikefly123:
Okay actually scratch what I previously said, I am now seeing the bug. Will continue working the problem!

nateinaction (approved):
Prepping for repo conversion. Will not be possible to revert PR after conversion. I intend to put up a new PR with the async changes.

Issue #203, opened 2025-03-21
Min-Max Ranges for Config Values


Alandlt15:
Hello team! I'm currently working on issue #128 and adding a new command so a mission operator can change values temporary or permanently at runtime. 

But we need to really make sure the new values are secure and do not harm the satellite, so I'm just looking to find the min and max range for the values that have -1 and 1; excluding bool types. (Michael provided me with some ranges but it was very improptu, so some might be incorrect):

<img width="469" alt="Image" src="https://github.com/user-attachments/assets/6dabdffe-31e7-4bba-828a-01ce123710da" />

<img width="422" alt="Image" src="https://github.com/user-attachments/assets/f0579ca4-f662-4bd2-878d-768a12d2fd08" />

Mikefly123:
Hey @Alandlt15! Here's a typed up list of the values you're still looking for:

- super_secret_code (cdh command password): type[bytes], min[1], max[24]
- repeat_code: type[bytes], min[1], max[4]
- last_battery_temp: Delete this, it should be in Jame's State of Health interface rather than config
- normal_charge_current: min[0.0], max[2000.0]
- normal_battery_voltage: min[6.0], max[8.4]
- critical_battery_voltage: min[5.4], max[7.2]
- battery_voltage: Move to State of Health Interface
- current_draw: Move to State of Health Interface
- sleep_duration: min[1], max[86400]
- normal_battery_temp: min[1], max[35]
- normal micro_temp: min[1], max[50]
- sender_id: min[0], max[255]
- start_time: min[0], max[80000]
- transmit_frequency: min0[435], max0[438], min1[915], max1[915]
- ack_delay: min[0.0], max[2.0]
- braodcast_address: min[0], max[255]
- node_address: min[0], max[255]
- modulation_type: min[0], max[1]

Alandlt15:
Thank you so much!

Issue #128, opened 2025-01-31
Minor Updates Through Overwriting `config.json`


Mikefly123:

Alandlt15:
@Mikefly123 I plan to start working on this over the break and hopefully I can get some good progress done. As this past Monday, at the lab we played around with the USBFunctions found in lib.pysquared.usb. Ultimately, we were able to create files in the sd card and even edit these files through the use of the usb functions. 

Later that day, we to also edited the current config.json file through the board from the repl which was really cool and amazing to see. Since we now know what steps we need to take for the board to write to its own file system, I'm starting to look at implementing the command to overwrite config.json and I have some questions about it.

1. What is the process of creating/adding a command and how do we hone in on a specific byte array for the command?
 
2. I'm not sure if this is a good question, but what is the encoding/decoding preference for our satellite? 

3. This one might be like question 2, but when I print this out for one of the commands `print(b"\xd4\x9f".decode("utf-8", errors="ignore"))` I get 'K' as output, and when I print this `print(int.from_bytes(b"\xd4\x9f", "big"))` I get 54431, I'm not sure if this matters or not?

Mikefly123:
Hey @Alandlt15 a quick answer to your questions:

1. To add a command, you simply add a function to the `cdh.py` that you would like to be executed when the bytecode is called in the commands dictionary. There is no system right now for defining the byte arrays, so you can use whatever you want. We probably should standardize command codes at some point in the future though.
2. There is currently no preference, but David and Madhav are working on a system right now for standardizing operational codes in our packet headers. You may want to catch up with their work!
3. No that doesn't matter, because the `cdh.py` is just checking if any two bytes are like any other two bytes they are not limited to ASCII characters. In that first example there happens to be an ASCII representation of that byte code, but in the second one you are getting the integer representation of the byte code (don't you love Python's mutability haha)

Alandlt15:
Awesome! Thank you for taking the time to answer I'll definitely get in touch with David and Madhav to see how their work is so far. 

Lol heck yea! Python being Python.

PR #210, opened 2025-03-25
Overwritting config values

Alandlt15:
## Summary
This PR focuses on the validation logic of overwriting config values. I plan on uploading another PR for the command part where I will add the command for overwriting config values to the cdh class, and I may have to change the interface for sending commands in radio_test.py. There's also another command that should return the current version and last modified date of the software and config files that I will work on. (#128). 

I'm only missing the min and max ranges for config values (#203).

Validation Overview:
1. Checks if the value wanting to be changed is actually allowed to be changed (Using schema's for this)
2. Checks that the value's type matches the expected type
3. Checks that the value is in a range (min >= & <= max)
4. Checks if its a temporary value or permanent change

## How was this tested
- [x] Added new unit tests
- [ ] Other (Please describe)

I have not tested if normal execution of the program runs fine with what I have, but I will test once I get access to my board lol. But the meat and potatoes will be when we confirm this fully works once we are able to send commands through radio.


Alandlt15:
Got full coverage in config:
image

Just missing to move 3 variables to the state of health interface that James is working on.

Alandlt15:
About to test on board

Alandlt15:
another pr is up for this

PR #213, opened 2025-03-26
Fix modulation value retrieval in state of health logging

JamesDCowley:
## Summary

The issue was that the `radio_manager.get_modulation()` can return a string literal `"LoRa"` or `"FSK"` which can't be casted into an int. I removed the int cast and that seemed to fix the issue.

## How was this tested
- [ ] Added new unit tests
Here's the same code change but on my State of Health branch (so that the results are clear)
image
image
On this branch, you can't see the results (because they are beaconed) but the base 10 error is no longer thrown.
- [ ] Other (Please describe)


blakejameson (approved):
Just ran the code on my board and no longer got hit with the Base 10 error
Good job James

PR #217, opened 2025-03-26
Using `make sync-time` to update the time on the board,

blakejameson:
## Summary
Currently, depending on what firmware loaded on the board, the default date is January 1, 2020 or January 1, 2000

The `make sync-time` command will update the time on the board to be the current real world time. The command is supported for Mac, Linux, and Windows

However, something to note is that the board resetting will cause the time to reset to the default. As we are currently using the RP2040's RTC and not the external RTC, the time is lost when either losing power or during a 'self.micro.reset' when the board is running for an hour.

## How was this tested
- [ ] Added new unit tests

Screenshot 2025-03-26 at 5 41 12 PM
Screenshot 2025-03-26 at 5 41 17 PM
Screenshot 2025-03-26 at 5 41 34 PM

- [ ] Other (Please describe)


blakejameson:
Closing this, as sync-time will instead live in https://github.com/proveskit/sync-time

PR #223, opened 2025-03-29
Remove unused items and extract inits from Satellite class

nateinaction:
## Summary
Satellite class becomes ~150 lines after this change. Removed many options from the Satellite class like `legacy` which was always false or `orpheus` which was always true. This also removed TCA Multiplexer implementation since it seemed like something we could reimplement after this change and was only used with faces which are rarely used in the lab right now. Removed Neopixel implementation because it also seems like something we could easily readd. Also removed sdcard init since there is something equivalent in the usbfunctions test package and is still experimental. 

Added a gazillion new tests.

## How was this tested
- [x] Added new unit tests
- [ ] Other (Please describe)

Ran on v5 board:
Screenshot 2025-03-31 at 19 15 22


Mikefly123 (approved):
LGTM!

Meme of the Review: https://youtu.be/XE1XE6Thp9E?feature=shared

hrfarmer (approved):

PR #228, opened 2025-04-07
Adding functionality to serialize invalid types as strings

blakejameson:
## Summary
In recent time, Nate pointed out in https://github.com/proveskit/pysquared/issues/194 that CircuitPython 9,2,1+ "Raises TypeError when attempt to produce JSON for invalid types."

He added code in https://github.com/proveskit/pysquared/pull/206 that gracefully handled TypeError.

I had an idea to detect if an invalid object occurred as a kwarg trying to be parsed and logged, and cast that object to a string if invalid. This results in being able to log the output we had before the changes.

As seen below, the 'pin' kwarg is able to be passed to the logger again and gives descriptive info during the "Initializing pin" and "Initializing watchdog" logs. The previous result would have been a log that said "Failed to serialize log message" 

Screenshot 2025-04-07 at 6 12 24 PM

## How was this tested
- [x] Added new unit tests
- [ ] Other (Please describe)

### DIFF
--- a/pysquared/hardware/digitalio.py
+++ b/pysquared/hardware/digitalio.py
@@ -21,10 +21,7 @@ def initialize_pin(
 
     :return ~digitalio.DigitalInOut: The initialized DigitalInOut object.
     """
-    logger.debug(
-        message="Initializing pin",
-        initial_value=initial_value,
-    )
+    logger.debug(message="Initializing pin", initial_value=initial_value, pin=pin)
 
     try:
         digital_in_out = DigitalInOut(pin)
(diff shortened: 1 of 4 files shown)


Mikefly123 (approved):
LGTM! Thanks for the great work

blakejameson:
@Mikefly123 Thanks Michael! Something I just noticed is I forgot to put an underscore for the function I introduced, `is_valid_json_type`. Following the current convention of putting an underscore for a private method (only Logger class uses this method), I am going to go ahead and change this

Mikefly123 (approved):
Good catch @blakejameson LGTM again!

PR #230, opened 2025-04-12
[ADR] Safe Device Failure

nateinaction:
## Summary
This pull request introduces a new Architectural Decision Record (ADR) for handling safe device failures on the satellite. The ADR proposes patterns to replace the now removed `hardware` dictionary that used to track device initialization state.

### DIFF
--- a/adr/safe-device-failure.md
+++ b/adr/safe-device-failure.md
@@ -1,0 +1,198 @@
+# Safe Device Failure
+
+## Status
+
+Proposed
+
+## Context
+
+This ADR proposes patterns allowing non-critical devices to fail while ensuring downstream services are not impacted. For example, if the IMU fails to initialize, we want to ensure that the satellite continues operation, but services such as de-tumbling can understand that the IMU is not available and adjust their behavior accordingly. In this case, the de-tumbling service would not engage magnetorquers, reaction wheels, or thrusters.
+
+This ADR does not attempt to solve partial failures of a device where only some of its functionality is available or when it returns faulty data. This problem can be addressed in a future ADR.
+
+### Prior Implementation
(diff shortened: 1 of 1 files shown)


Mikefly123:
Hey @nateinaction! Thanks for putting this together and this is a great way to document decision points going forward. I added a tag that can be used for these kinds of issues so we can go back to them in the future easily.

After spending some time thinking about the two options, I think **Option #2** would be preferable to try and maintain the highest level of code safety without putting too much of a burden on the end user. Especially because the expectation is that this codebase will primarily be used by beginner software teams with very limited resources, creating a system where it is unlikely or impossible that downstream function calls might become an unhandled exception minefield feels like it should be a top priority.

In my mind one of the other potential benefits of going with the second option is making it easier to implement a ["Circuit Breaker Pattern"](https://learn.microsoft.com/en-us/azure/architecture/patterns/circuit-breaker) type of thing to allow downstream services to still call into *something* which has built in functionality to try and restore the connection to the downstream sensor over time.

adr/graceful-device-initialization-failure-handling.md
```diff
@@ -0,0 +1,198 @@
...
+### Current State
+Currently, all device initialization failures are fatal, meaning that if a device fails to initialize, the entire root `try/except` in `main.py` will be triggered, and the satellite will restart. For example, the [IMU initialization](https://github.com/proveskit/CircuitPython_RP2040_v4/blob/4693e43a373030e74dad109c0858c9e36998f1df/main.py#L94) in the v4 board repo is not handling exceptions.
+
+Possibly by accident, all current hardware device managers are safe to call after a failed initialization. As long as their initialization is wrapped in a `try/except` block, the device manager will return `None` for all subsequent calls for sensor readings.
+
+## Proposal
+
+### Pattern 1: Generic Device Protocol with `get_state()`
+
+We propose to add a `get_state()` method to the generic device protocol. This method will return 1 of 3 possible states: `unknown`, `initialized`, or `failed` allowing downstream services to check the state of a device before attempting to use it.
```
nateinaction, line 25:
```suggestion
We propose to add a `get_state()` method to the generic device protocol. This method will return 1 of 3 possible states: `disabled`, `enabled`, or `failed` allowing downstream services to check the state of a device before attempting to use it.
```

PR #231, opened 2025-04-12
Exclude protocols from code coverage reporting

nateinaction:
## Summary
Functions defined by protocols are run directly and only contain a `pass` or `...` statement. Code coverage is complaining that we don't run them so let's exclude them.

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)


JamesDCowley (approved):
LGTM!

PR #235, opened 2025-04-17
New State of Health

### LINKED ISSUE #159: [TASK] Create a `state_of_health` Interface to Control Power Modes
## Story
* **As a** `circuitry_flight_software` developer.
* **I want** a simpler way to interact with the satellite's state_of_health.
* **So that** we have a single source of truth of the current health of the satellite systems to determine which mode we should be in.

The `state of health` interface of the satellite should serve two roles, collecting metadata (like temperatures of key components and the voltage of the battery) and changing the current operational mode of the satellite as a result of what this meta data tells us. Implementing a `mode_manager` should probably be a separate task. 

## Data Points to Collect
All of these data points should be collected by the module and saved to a single data structure.
- [ ] System Voltage 
- [ ] System Current
- [ ] Solar Voltage
- [ ] Solar Current
- [ ] Radio Temperature
- [ ] Microcontroller Temperature
- [ ] Internal Temperature
- [ ] Error Count

> NOTE: If you are currently testing on a V4 flight controller board without a battery board there is no way to collect the voltage and current information. This might be a bit of a blocker if there is no easy way to just ignore the fact that those sensors are not connected.

## Acceptance Criteria

- [ ] Clean Out All of the Old State of Health Functions and Attributes
- [ ] Consolidate state of health collection into a single interface module
- [ ] Implement unit tests on the state of health system (example: checking that mode switching is done correctly)

## Technical Details
A great place to start with this is in `pysquared.py`, where we have this function that sets the `power_mode` and manipulates some of the hardware as a result of the current `mode`: 

```py 
    def powermode(self, mode: str) -> None:
        """
        Configure the hardware for minimum or normal power consumption
        Add custom modes for mission-specific control
        """
        try:
            if "crit" in mode:
                self.neopixel.brightness = 0
                self.enable_rf.value = False
                self.power_mode: str = "critical"

            elif "min" in mode:
                self.neopixel.brightness = 0
                self.enable_rf.value = False

                self.power_mode: str = "minimum"

            elif "norm" in mode:
                self.enable_rf.value = True
                self.power_mode: str = "normal"
                # don't forget to reconfigure radios, gps, etc...

            elif "max" in mode:
                self.enable_rf.value = True
                self.power_mode: str = "maximum"
        except Exception as e:
            self.logger.error(
                "There was an Error in changing operations of powermode",
                err=e,
                mode=mode,
            )
```
The `power_mode` attribute is then checked in `main.py` to determine what functional flow the satellite should take at any given time:
```py
    ######################### MAIN LOOP ##############################
    try:
        while True:
            # L0 automatic tasks no matter the battery level
            c.check_reboot()

            if c.power_mode == "critical":
                c.RGB = (0, 0, 0)
                critical_power_operations()

            elif c.power_mode == "minimum":
                c.RGB = (255, 0, 0)
                minimum_power_operations()

            elif c.power_mode == "normal":
                c.RGB = (255, 255, 0)
                main()

            elif c.power_mode == "maximum":
                c.RGB = (0, 255, 0)
                main()

            else:
                f.listen()
```

JamesDCowley:
## Summary
Closes #159 (finally). Created a new State of Health class to hold information from a bunch of sensors such as temperatures, voltage, current, etc. Will be used for upcoming Mode Manager to manage power states.
## How was this tested
- [X] Added new unit tests
image
(Power monitor stuff is null but does work with a working battery board, but ours blew up)
Works on v4 flight boards with v2 battery boards. Make sure to import and add the `INA219Manager` and `StateOfHealth` into `main.py` on whatever version repo.

`main.py` For v4 flight + v2 battery (will be added in upcoming PR to v4 and v5 boards):
```python
from lib.pysquared.hardware.power_monitor.manager.ina219 import INA219Manager
...
from lib.pysquared.state_of_health import StateOfHealth
...
# v4 only -- use i2c1 for v5
i2c0 = initialize_i2c_bus(
        logger,
        board.I2C0_SCL,
        board.I2C0_SDA,
        100000,
)
...
battery_power_monitor = INA219Manager(logger, i2c0, 0x40) # address and bus may be different depending on your board
soh = StateOfHealth(
        logger,
        battery_power_monitor,
        solar_power_monitor,
        radio,
        imu,
        Counter(index=register.BOOTCNT, datastore=microcontroller.nvm), # boot count
        Flag(index=register.FLAG, bit_index=6, datastore=microcontroller.nvm), # burned
        Flag(index=register.FLAG, bit_index=3, datastore=microcontroller.nvm), # brownout
        Flag(index=register.FLAG, bit_index=7, datastore=microcontroller.nvm), # fsk
)
# you can put None in place of the power monitors if you don't have a battery board and are using a v4 flight board
...
def main():
...
        state_of_health.update()
...
```
- [ ] Other (Please describe)

### DIFF
--- a/pysquared/power_health.py
+++ b/pysquared/power_health.py
@@ -51,18 +51,18 @@ class PowerHealth:
         try:
             bus_voltage = self._avg_reading(self._power_monitor.get_bus_voltage)
             if bus_voltage is None:
-                self.logger.error("Power monitor failed to provide bus voltage reading")
+                self.logger.warning(
+                    "Power monitor failed to provide bus voltage reading"
+                )
                 return UNKNOWN()
 
             current = self._avg_reading(self._power_monitor.get_current)
             if current is None:
-                self.logger.error("Power monitor failed to provide current reading")
+                self.logger.warning("Power monitor failed to provide current reading")
                 return UNKNOWN()
 
         except Exception as e:
-            self.logger.error(
-                "Exception occurred while reading from power monitor", error=e
-            )
+            self.logger.error("Exception occurred while reading from power monitor", e)
             return UNKNOWN()
 
         # Critical check first - if battery voltage is below critical threshold
--- a/tests/unit/test_power_health.py
+++ b/tests/unit/test_power_health.py
@@ -188,7 +188,7 @@ def test_get_with_none_voltage_reading(power_health):
     result = power_health.get()
 
     assert isinstance(result, UNKNOWN)
-    power_health.logger.error.assert_called_with(
+    power_health.logger.warning.assert_called_with(
         "Power monitor failed to provide bus voltage reading"
     )
 
@@ -204,7 +204,7 @@ def test_get_with_none_current_reading(power_health):
     result = power_health.get()
 
     assert isinstance(result, UNKNOWN)
-    power_health.logger.error.assert_called_with(
+    power_health.logger.warning.assert_called_with(
         "Power monitor failed to provide current reading"
     )
 
@@ -218,12 +218,10 @@ def test_get_with_exception_during_voltage_reading(power_health):
     result = power_health.get()
 
     assert isinstance(result, UNKNOWN)
-    # Check that error was called with the correct message and exception type/message
-    power_health.logger.error.assert_called_once()
-    call_args = power_health.logger.error.call_args
-    assert call_args[0][0] == "Exception occurred while reading from power monitor"
-    assert isinstance(call_args[1]["error"], RuntimeError)
-    assert str(call_args[1]["error"]) == "Sensor communication error"
+    # Check that error was called with the correct message and exception as positional parameter
+    power_health.logger.error.assert_called_once_with(
+        "Exception occurred while reading from power monitor", test_exception
+    )
 
 
 def test_get_with_exception_during_current_reading(power_health):
@@ -235,12 +233,10 @@ def test_get_with_exception_during_current_reading(power_health):
     result = power_health.get()
 
     assert isinstance(result, UNKNOWN)
-    # Check that error was called with the correct message and exception type/message
-    power_health.logger.error.assert_called_once()
-    call_args = power_health.logger.error.call_args
-    assert call_args[0][0] == "Exception occurred while reading from power monitor"
-    assert isinstance(call_args[1]["error"], RuntimeError)
-    assert str(call_args[1]["error"]) == "Current sensor failed"
+    # Check that error was called with the correct message and exception as positional parameter
+    power_health.logger.error.assert_called_once_with(
+        "Exception occurred while reading from power monitor", test_exception
+    )
 
 
 def test_get_with_sensor_method_exception(power_health):
@@ -252,12 +248,10 @@ def test_get_with_sensor_method_exception(power_health):
     result = power_health.get()
 
     assert isinstance(result, UNKNOWN)
-    # Check that error was called with the correct message and exception type/message
-    power_health.logger.error.assert_called_once()
-    call_args = power_health.logger.error.call_args
-    assert call_args[0][0] == "Exception occurred while reading from power monitor"
-    assert isinstance(call_args[1]["error"], OSError)
-    assert str(call_args[1]["error"]) == "I2C communication failed"
+    # Check that error was called with the correct message and exception as positional parameter
+    power_health.logger.error.assert_called_once_with(
+        "Exception occurred while reading from power monitor", test_exception
+    )
 
 
 def test_get_logs_sensor_debug_info(power_health):


pysquared/functions.py
nateinaction, line 1: I like that you're removing things from `functions.py`. Thank you!

pysquared/state_of_health.py
```diff
@@ -0,0 +1,174 @@
...
+                ("brownout_flag", None),
+                ("fsk_flag", None),
+            ]
+        )
+
+    def update(self):
+        """
+        Update the state of health
+        """
+        try:
```
nateinaction, line 67: Looking over the calls in this try/except I think that none of these calls have the possibility of raising an exception. What are you seeing?

pysquared/state_of_health.py
```diff
@@ -0,0 +1,174 @@
...
+        :return: The battery voltage in volts
+        :rtype: float | None
+
+        """
+        if self.battery_power_monitor is not None:
+            voltage: float = 0
+            try:
+                for _ in range(50):
+                    voltage += self.battery_power_monitor.get_bus_voltage()
+                return voltage / 50 + 0.2  # volts and correction factor
```
nateinaction, line 121: Where does this correction factor come from?

pysquared/state_of_health.py
```diff
@@ -0,0 +1,174 @@
...
+        :rtype: float | None
+
+        """
+        if self.battery_power_monitor is not None:
+            voltage: float = 0
+            try:
+                for _ in range(50):
+                    voltage += (
+                        self.battery_power_monitor.get_bus_voltage()
+                        + self.battery_power_monitor.get_shunt_voltage()
```
nateinaction, line 102: Where does this idea for adding these two voltages together come from?

nateinaction (commented):
This is a pretty great refactor from the old state of health. Thank you for extracting this functionality from `functions.py`. This feels like the start of a new "services" layer in `pysquared`.

While you were still building out the ina219 power monitor manager I remember us talking about how useful the protos could be in creating a generic state of health service. While not as customized as _this_ implementation, in #240 I took a swing at spinning up a generic state of health service just to see what it could look like. How do you think of the generic model compares to what you have here?

pysquared/state_of_health.py
Mikefly123: This was based on the legacy implementation that we took from Max Holliday's PyCubed! 

Good flag on this Nate, because looking at it again I realized that the implementation of `system_voltage()` and `battery_voltage()` is incorrect. The legacy PyCubed implementation uses an `ADM1176` power monitor chip whereas our PySquared boards use the `INA219` power monitor chip.

This code should actually be changed so the `bus_voltage() + shunt_voltage()` call is in the `battery_voltage()` instead. This is explained a bit in Adafruit's guide for the [INA219](https://learn.adafruit.com/adafruit-ina219-current-sensor-breakout/python-circuitpython):
```py
    # INA219 measure bus voltage on the load side. So PSU voltage = bus_voltage + shunt_voltage
    print("Voltage (VIN+) : {:6.3f}   V".format(bus_voltage + shunt_voltage))
```
`system_voltage()` should just be based on `bus_voltage`, which is the actual amount of voltage available to the system vs what the voltage of the batteries is.

Mikefly123 (changes requested):
Noticed based on Nate's comments that there should be a slight change to the implementation of the `system_voltage()` and `battery_voltage()` calls!

Mikefly123:
@nateinaction I think it could go either way with keeping the temperature monitoring in here for flavor or splitting it off into a dedicated thermal manager / interface. It is true that right now there is not actually much we can do in response to our of range temperatures and it is mostly just a nice FYI for the ground operators.

If you'd like I can see if someone on my team has some time to address the `test` and `typecheck` errors so we can more forward with this.

Mikefly123:
Hey @yudataguy I think the updates look good, a few things to take a look at though:

1. It looks like two of your tests are failing, one because of the float =/= int comparison thing and the other due to a `__name__` attribute error. If you can straighten these out that would be great! 

2. Rather than using an absolute value threshold based on `self.config.normal_battery_voltage - self.config.critical_battery_voltage` I think it would be more clear if we just added another line to `config` for `degraded_battery_voltage = 7.0V or something like that`. So if it goes below that number then we are in the degraded range.

pysquared/state_of_health.py
```diff
@@ -0,0 +1,100 @@
...
+class PowerHealth:
+    def __init__(
+        self,
+        logger: Logger,
+        config: Config,
+        power_monitor: PowerMonitorProto,
+    ) -> None:
+        self.logger: Logger = logger
+        self.config: Config = config
+        self._sensor: PowerMonitorProto = power_monitor
```
nateinaction, line 37: Maybe rename since there is only one?
```suggestion
        self._power_monitor: PowerMonitorProto = power_monitor
```

pysquared/state_of_health.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def get(self) -> NOMINAL | DEGRADED | CRITICAL:
+        """
+        Get the power health
+        """
+        errors: List[str] = []
+        self.logger.debug("Power monitor: ", sensor=self._sensor)
+
+        if isinstance(self._sensor, PowerMonitorProto):
```
nateinaction, line 46: No longer need the isinstance check since we know `self._sensor` is of type `PowerMonitorProto`

pysquared/state_of_health.py
nateinaction, line 1: Should we change the file name to `power_health.py` to match the test and name of class?

nateinaction (commented):
Looking good Sam! I had a few small comments about naming and one about removing a logic block that is no longer necessary with your refactor. Great job!

PR #242, opened 2025-04-28
Create sx1280 manager

nateinaction:
## Summary
This pull request introduces support for the SX1280 radio module by adding a mock implementation, updating dependencies, and creating a new manager class for handling SX1280 radios.

## How was this tested
- [x] Added new unit tests
- [ ] Other (Please describe)

This has been tested on the v5a development branch but I don't have screenshots or a board on me atm.

### DIFF
--- a/pysquared/hardware/radio/manager/sx1280.py
+++ b/pysquared/hardware/radio/manager/sx1280.py
@@ -1,0 +1,97 @@
+from busio import SPI
+from digitalio import DigitalInOut
+from sx1280.sx1280 import SX1280
+
+from ....config.radio import RadioConfig
+from ....logger import Logger
+from ....nvm.flag import Flag
+from ..modulation import LoRa, RadioModulation
+from .base import BaseRadioManager
+
+# Type hinting only
+try:
+    from typing import Optional, Type
+except ImportError:
+    pass
+
+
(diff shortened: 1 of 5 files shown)


pysquared/hardware/radio/manager/sx1280.py
```diff
@@ -0,0 +1,101 @@
...
+class SX1280Manager(BaseRadioManager):
+    """Manager class implementing RadioProto for SX1280 radios."""
+
+    _radio: SX1280
+
+    def __init__(
+        self,
+        logger: Logger,
+        radio_config: RadioConfig,
+        use_fsk: Flag,
```
Mikefly123, line 27: I think this links in with the thoughts from #244, but we are in a bit of a weird spot right now with only one unified config (which was all that was needed for V4 boards). With two radios on V5x boards I think it is fine to just used similar settings for both radios for now. In the near future though we'll probably want to find a good way to split the configs so you can have both radios doing different things.

As you were saying on the other PR, perhaps for now we just cut the 2.X.X releases with more simple and straight forward config and in a 3.X.X release we can expand with more protocols.

Mikefly123 (approved):
This looks great to me Nate! Thanks for the hard work!

PR #244, opened 2025-05-07
Add modify_config function

hrfarmer:
## Summary
This adds a `modify_config` function to the radio manager base class, which takes in a `RadioConfig` class, sets the passed in class as the manager's config, and then re-initializes the radio.

## How was this tested
- [x] Added new unit tests
image

(Taken in the REPL)

### DIFF
--- a/pysquared/hardware/radio/manager/base.py
+++ b/pysquared/hardware/radio/manager/base.py
@@ -110,6 +110,11 @@ class BaseRadioManager(RadioProto):
                 current=current_modulation,
             )
 
+    def modify_config(self, radio_config: RadioConfig) -> None:
+        """Modify the radio configuration. This will apply the new configuration and reinitialize the radio."""
+        self._radio_config = radio_config
+        self._initialize_radio(self.get_modulation())
+
     # Methods to be overridden by subclasses
     def _initialize_radio(self, modulation: Type[RadioModulation]) -> None:
         """Initialize the specific radio hardware.
--- a/pysquared/hardware/radio/manager/rfm9x.py
+++ b/pysquared/hardware/radio/manager/rfm9x.py
@@ -174,3 +174,30 @@ class RFM9xManager(BaseRadioManager, TemperatureSensorProto):
         except Exception as e:
             self._log.error("Error receiving data", e)
             return None
+
+    def modify_config(self, **kwargs) -> None:
+        """Modify the radio configuration.
+
+        :param kwargs: Configuration parameters to modify.
+        """
+        if kwargs["sender_id"] is not None:
+            # verify is string?
+            self._radio.node = kwargs["sender_id"]
+            self._log.debug(
+                "Radio sender ID modified",
+                sender_id=kwargs["sender_id"],
+            )
+
+        if self.get_modulation() == FSK:
+            if kwargs["broadcast_address"] is not None:
+                self._radio.fsk_broadcast_address = kwargs["broadcast_address"]
+                self._log.debug(
+                    "Radio broadcast address modified",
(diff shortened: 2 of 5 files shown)


pysquared/hardware/radio/manager/base.py
```diff
@@ -110,6 +110,11 @@ def set_modulation(self, modulation: Type[RadioModulation]) -> None:
                 current=current_modulation,
             )
 
+    def modify_config(self, radio_config: RadioConfig) -> None:
+        """Modify the radio configuration. This will apply the new configuration and reinitialize the radio."""
+        self._radio_config = radio_config
+        self._initialize_radio(self.get_modulation())
```
nateinaction, line 116: We may be able to loosen these rules now that we have more memory but, as of right now, we're not allowing the radio to be reinitialized during runtime. Doing so makes understanding the memory state of the satellite more challenging.

Instead my recommendation is to implement these in the individual radio managers and allow them to modify any variables that we are allowed to set outside of initialization.

pysquared/protos/radio.py
```diff
@@ -45,3 +46,10 @@ def receive(self, timeout: Optional[int] = None) -> Optional[bytes]:
         :rtype: Optional[bytes]
         """
         ...
+
+    def modify_config(self, radio_config: RadioConfig) -> None:
```
nateinaction, line 50: Taking in `RadioConfig` is a little concerning.

I'm concerned, in general, about config in pysquared right now. It has a lot of embedded business logic that probably shouldn't be there. For example, changing node and destination is only valid for the RFM9x radio library when running in LoRa mode. We are supporting other radios and those all have to share the same config.

I think we should probably just go with the config management that we have today for 2.x and try to address these concerns in a future 3.x release. Michael suggested having a config for each manager? Maybe that would be a better way to organize?

Regardless, I think taking in `RadioConfig` is OK for this release.

nateinaction (changes requested):
Awesome to see a working solution! Great job!

General scaffold is good, only requested change is around reinitializing radio.

pysquared/hardware/radio/manager/rfm9x.py
```diff
@@ -80,6 +80,59 @@ def _send_internal(self, payload: bytes) -> bool:
...
+        """Modify the radio configuration. This will apply any new configuration options during runtime."""
+
+        # Validate all radio config parameters
+        self._radio_config.validate("sender_id", radio_config.sender_id)
+        self._radio_config.validate("receiver_id", radio_config.receiver_id)
+        self._radio_config.validate(
+            "transmit_frequency", radio_config.transmit_frequency
+        )
+        self._radio_config.validate("start_time", radio_config.start_time)
+        self._radio_config.validate("license", radio_config.license)
```
nateinaction, line 93:
```suggestion
```

tests/unit/hardware/radio/manager/test_rfm9x_manager.py
```diff
@@ -746,3 +746,48 @@ def test_receive_exception(
...
+    mock_radio_config: RadioConfig,
+    mock_use_fsk: MagicMock,
+):
+    """Test modifying the radio configuration."""
+    mock_use_fsk.get.return_value = False
+    mock_radio_instance = MagicMock(spec=RFM9x)
+    mock_radio_instance.receive = MagicMock()
+    receive_error = RuntimeError("Receive Error")
+    mock_radio_instance.receive.side_effect = receive_error
+    mock_rfm9x.return_value = mock_radio_instance
```
nateinaction:
```suggestion
```

nateinaction (approved):

nateinaction (approved):

PR #254, opened 2025-05-27
Get radio modulation from config

hrfarmer:
## Summary
Switches to getting the radio modulation from config.json instead of using the nvm flag on the board. Also removes `set_modulation()` from the radio manager(s) since that should now be handled by `update_config()` in the Config class.

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)


pysquared/config/radio.py
```diff
@@ -17,6 +18,7 @@ def __init__(self, radio_dict: dict) -> None:
 
         self.RADIO_SCHEMA = {
             "license": {"type": str},
+            "modulation": {"type": str},
```
nateinaction, line 21: Would be nice to specify which values are allowed here so folks can't set incorrect values in config.

pysquared/config/radio.py
```diff
@@ -8,6 +8,7 @@
 class RadioConfig:
     def __init__(self, radio_dict: dict) -> None:
         self.license: str = radio_dict["license"]
+        self.modulation: str = radio_dict["modulation"]
```
nateinaction, line 11:
```suggestion
        self.modulation: Literal["LoRa", "FSK"] = radio_dict["modulation"]
```

pysquared/hardware/radio/manager/base.py
```diff
@@ -99,16 +98,12 @@ def receive(self, timeout: Optional[int] = None) -> bytes | None:
...
-                "Radio modulation change requested for next init",
-                requested=modulation,
-                current=current_modulation,
-            )
+    def get_modulation(self) -> Type[RadioModulation]:
+        """Get the modulation mode from the initialized radio hardware.
+
+        :return: The current modulation mode of the hardware.
+        """
+        return FSK if self._radio_config.modulation == "FSK" else LoRa
```
nateinaction, line 106: Is this always true though? In theory we should be able to modify the config in flight while the radio is set to a different modulation. Did you prefer this implementation over the previous per-manager implementations?

pysquared/hardware/radio/manager/rfm9x.py
```diff
@@ -125,10 +122,6 @@ def modify_config(self, radio_config: RadioConfig) -> None:
                 self._radio.preamble_length = self._radio.spreading_factor  # type: ignore
                 self._radio.low_datarate_optimize = 1  # type: ignore
 
-    def get_modulation(self) -> Type[FSK] | Type[LoRa]:
-        """Get the modulation mode from the initialized RFM9x radio."""
-        return FSK if self._radio.__class__.__name__ == "RFM9xFSK" else LoRa
-
```
nateinaction, line 131: Same Q

pysquared/hardware/radio/manager/sx126x.py
```diff
@@ -78,10 +72,6 @@ def _send_internal(self, payload: bytes) -> bool:
             return False
         return True
 
-    def get_modulation(self) -> Type[RadioModulation]:
-        """Get the modulation mode from the initialized SX126x radio."""
-        return FSK if self._radio.radio_modulation == "FSK" else LoRa
-
```
nateinaction, line 84: Same Q

nateinaction (commented):
Love seeing the flag gone! Great work on this!

pysquared/hardware/radio/manager/base.py
Mikefly123: @hrfarmer I think using this if/else statement in the base class is alright for this version of the flight software. We might want to sprinkle a TODO statement here to revisit this though as we explore manipulating the modulation properties of the underlying radio drivers some more. 

Right now, it is true we only directly support FSK and LoRa as options, but in the future there might be changes to the software that allow us to use GMSK (a subset of FSK) or BPSK (in the case of the SX1280). These updates are not necessarily on the roadmap right now though, so keeping this part of the API more understandable and maybe changing the behavior later sounds okay to me.

Open to other thoughts, but I think this will be the last thing for us to resolve with this PR before merging it in!

Issue #249, opened 2025-05-14
[FEATURE] Burnwire Manager


Mikefly123:
## Story
* **As a** mission operator
* **I want** to have a way of firing the burn wire
* **So that** we can safely and reliably deploy the satellite's antennas.

## Acceptance Criteria

- [x] Accepts a set of initialized pins that enable and disable the burn wire
- [x] Implements a `burn` function that activates the burn wire for a set amount of time
- [ ] #262

## Technical Details
This will be a tape out of the initial reimplementation of the burn wire functionalities for the `pysquared` V2.0.0 release. The burn wire circuit is very simple, as seen in the following schematic (from V5b Flight Controller Board) it is a pair of load switches that are cascaded together in series.

![Image](https://github.com/user-attachments/assets/1043f7b9-d75b-4760-b3b7-065acfb3cd33)

In the future we will ideally make a follow up to this that adds multiple retries and some kind of lockout functionality in case of erroneous behavior.

Mikefly123:
Started working on this on the new `burnwire-manager` branch! Initial takeout is almost done. 

I think I will actually take the last criteria for the smart_burn function here and split it off into its own thing. Pasting my placeholder for it here in case I want it later:

```py
    def smart_burn(self, max_retries: int = 3, timeout_duration: float = 5.0):
        """Fires the burnwire and uses a deployment sensor

        :param int max_retries: The maximum number of times the burnwire is allowed to retry before exitng.
        :param float timeout_duration: The max time to keep the burnwire on for if the deployment sensor doesn't detect deployment.

        :return: A Boolean indicating whether the burn occured sucessfully
        :rtype: bool

        :raises Exception: If there is an error toggling the burnwire pins.
        """

        self._log.debug(
            "smart_burn() has not been implemented yet. Use burn() for now..."
        )
        raise NotImplementedError
```

Mikefly123:
Closing this as mostly complete with #258 acting as the follow on task!

PR #258, opened 2025-06-03
Introducing a Burnwire Manager

Mikefly123:
## Summary
Hey team! While I was on the plane to my vacation this week I took some time to scaffold a burn wire manger to address #249! 

I didn't bring a board with me, so I haven't been able to test it on hardware yet. I did have the AI overlords code up some unit tests though, which actually did catch some potential logical issues with the code before making this PR.

### Intended Usage
I think the docstrings should explain all, but generally what you do with this is for every burn wire port you might want to use you would first initialize the pins you want to use separately as digitalio (catch here is you have to absolutely make sure to default initialize them to the `off` state haha) then pass those two pins in when you initialize the individual manager. You also pass in the logger and an optional param for setting whether the enable logic is `True` or `False`.

> NOTE: Burnwire and load switch enable logic will be `True` (the default) on V5b+ flight controller boards. On V5dev and V5a it is enable `False` which is quite confusing lol.

There is also a burn wire proto introduced with this PR. I was originally going to include the `smart_burn` (brunwire with a sensor for feedback on when to stop!) but that was going to convolute the proto because sometimes you won't have a sensor to pass into the manager. So this PR will be for the `dumb_burn` which just takes a `max_timeout` as a parameter and a future PR will introduce the `smart_burn` as its own manager class.

## How was this tested
- [X] Added new unit tests
- [ ] Other (Please describe)

Screenshot 2025-06-04 at 12 37 18 AM

### DIFF
--- a/pysquared/hardware/burnwire/manager/burnwire.py
+++ b/pysquared/hardware/burnwire/manager/burnwire.py
@@ -11,7 +11,7 @@ Usage Example:
 from lib.pysquared.hardware.burnwire.manager.burnwire import BurnwireManager
 ...
 
-antenna_deployment = BurnwireManger(logger, board.FIRE_DEPLOY1A, board.FIRE_DEPLOY1B, enable_logic = False)
+antenna_deployment = BurnwireManager(logger, board.FIRE_DEPLOY1A, board.FIRE_DEPLOY1B, enable_logic = False)
 
 antenna_deployment.burn()
 """


tests/unit/hardware/test_burnwire.py
```diff
@@ -0,0 +1,140 @@
...
+        result = burnwire_manager.burn(timeout_duration=1.0, max_retries=3)
+
+        assert result is True
+        assert burnwire_manager.number_of_attempts == 3  # Should succeed on first try
+        burnwire_manager._log.warning.assert_called_once()  # Should warn about multiple retries
+
+
+def test_burn_with_invalid_retries(burnwire_manager):
+    """Test burnwire activation with invalid retry count."""
+    result = burnwire_manager.burn(max_retries=0)
```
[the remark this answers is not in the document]
Mikefly123: Addressed this with the latest commit.

pysquared/hardware/burnwire/manager/burnwire.py
```diff
@@ -0,0 +1,147 @@
...
+                    )
+                    return False
+
+            elif max_retries < 1:
+                self._log.warning("burnwire max_retries cannot be 0 or negative!")
+                raise ValueError
+
+            else:
+                self._log.warning(
+                    "Multiple Retries with base burnwire not recommended! Consider using smart_burnwire..."
```
nateinaction, line 83: Sounds like smart_burnwire doesn't exist yet, should probably adjust comment.

pysquared/hardware/burnwire/manager/burnwire.py
```diff
@@ -0,0 +1,147 @@
...
+            self._disable: bool = False
+        else:
+            self._disable: bool = True
+
+        self._enable_burn: DigitalInOut = enable_burn
+        self._fire_burn: DigitalInOut = fire_burn
+
+        self.number_of_attempts: int = 0
+
+    def burn(self, timeout_duration: float = 5.0, max_retries: int = 1) -> bool:
```
nateinaction, line 51:
```suggestion
    @with_retries(max_attempts=3, initial_delay=1)
    def burn(self, timeout_duration: float = 5.0) -> bool:
```

Could use the `with_retries` decorator for this functionality but I wonder how that would work with your safing logic. I guess it could depend on if you think this is a setting the satellite operator will want to manage or if you can take a "best guess." We've been taking a "best guess approach with other hardware initialization (always setting 3 tries so far)

pysquared/hardware/burnwire/manager/burnwire.py
```diff
@@ -0,0 +1,147 @@
...
+        :param Digitalio fire_burn: A pin used for enabling a specific burnwire port.
+        :param bool enable_logic: Boolean defining whether the burnwire load switches are enabled when True or False. Defaults to `True`.
+        """
+        self._log: Logger = logger
+        self._enable: bool = enable_logic
+
+        if enable_logic:
+            self._disable: bool = False
+        else:
+            self._disable: bool = True
```
nateinaction, line 44: It feels a little awkward to store `_enable` and `_disable` booleans which both represent the same value. My initial thought is anywhere we use `self._disable` we can just replace it with `!self._enable`. I do see why some might prefer something like this for readability. In that case we could make helper funcs that read from a single var:
```py
def __init__(self, enable_logic):
    self._enable_logic = enable_logic

def _enable(self):
    try:
        self._enable_burn.value = self._enable
    except Exception as e:
        raise RuntimeError("Failed to set enable_burn pin") from e

    time.sleep(0.1)  # Short pause to stabilize load switches

    # Burnwire becomes active
    try:
        self._fire_burn.value = self._enable_logic
    except Exception as e:
        raise RuntimeError("Failed to set fire_burn pin") from e
    
def _disable(self):
  try:
      self._fire_burn.value = !self._enable_logic
      self._enable_burn.value = !self._enable_logic
      self._log.info("Burnwire safed")
  except Exception as e:
      raise RuntimeError("Failed to safe burnwire pins") from e
```

nateinaction (commented):
Great work on this! I had some thoughts on enable/disable. There are also a few untested logic blocks that look testable! Check sonarqube to see them!

Mikefly123:
Just did a quick functional test on hardware and the functions work like a charm! I need to update the docs a bit though to clarify how this is supposed to be used.

https://youtube.com/shorts/HL30mOByLnc?feature=share

pysquared/hardware/burnwire/manager/burnwire.py
Mikefly123: I was originally going to include smart_burnwire in this pr, but I realized it was possible a little messy with needing to pass in a sensor to the manager. I want to spend more time thinking about it, potentially creating an entirely separate manager class that can be instantiated so the user clearly knows the difference between just doing a pure time based burn wire and one that is using sensor feedback to determine when to stop

nateinaction: I think multiple managers sounds like the right approach. 🚀

wrong-formatt (approved):

PR #271, opened 2025-07-12
100% docs coverage, new getting started guide

nateinaction:
## Summary
Hey team, here is a draft of the new getting started, contributing, design, and radio test guides using our new mkdocs site. Additionally all code is now covered by documentation and a linter step has been added to ensure we always cover our code 100%. This will help new devs get associated with the codebase and LLMs should be able to read this information and better support AI assisted development.

Most of the descriptions of the python code were written by AI, there may be errors but I think most of it is correct. Big PR but worst case, some comments are incorrect. We should fix forward if we find issues.

I also removed the retry decorator. It felt like it was never helpful. If we do bring it back we should understand why we need retries and make sure the retry system we build meet our needs.
## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)


docs/design-guide.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Documentation
+
+We use [MkDocs](https://www.mkdocs.org/) to build our documentation. We write our documentation in Markdown, which is a lightweight markup language that is easy to read and write. We document our code using docstrings, which are special comments that describe the purpose and usage of a function or class. We also use type hints in our docstrings to provide additional information about the expected types of parameters and return values.
+
+When documenting your code, use clear and concise examples that demonstrate real-world usage. Here are improved templates for module, class, and function documentation:
+
+TODO(nateinaction): Look at ruff's docstring formatting: https://docs.astral.sh/ruff/formatter/#docstring-formatting
```
ineskhou, line 51: TODO left in, I'm assuming the links are there for users but just checking in

docs/design-guide.md
```diff
@@ -0,0 +1,213 @@
...
+    ValueError: If the reading value is out of expected range.
+"""
+```
+
+## Sensor Readings
+
+All sensor readings must be in SI units and stored in a structure that includes the time of the reading. Including the time of the reading is important for analysing sensor data and ensuring that processes such as detumbling and attitude control can be performed accurately.
+
+The following table lists possible sensor properties, their corresponding types and units for common sensor readings. The table was pulled directly from the [CircuitPython Design Guide](https://docs.circuitpython.org/en/latest/docs/design_guide.html#sensor-properties-and-units):
+
```
ineskhou, line 118: I actually love this so much! Was working on code that dealt with sensors values so ill def be looking back at this PR + documentation 10/10

docs/getting-started.md
```diff
@@ -0,0 +1,177 @@
...
+    1. **Install Required Packages**: Open a command prompt or Git Bash terminal and run the following command to install required packages:
+        ```sh
+        choco install make rsync zip
+        ```
+
+    Keep in mind that the rest of this guide expects that you are using Git Bash.
+
+??? note "WSL Guide"
+    Follow the steps in the Linux Guide
+
```
ineskhou, line 44: You use screen and putty which makes sense! But in my tutorial I use tabby  

https://docs.google.com/document/d/1NCNVeGEz3Z3D182ieqnY65_4rggvdM3W5b1FYeuR4n4/edit?tab=t.0#heading=h.v9o48dk7c2mu

Thoughts on leaving both in as options in both tutorials? You can just copy what I have in the Part 2: Software Installation

(also the docs is temporary I'll put it all in a docs PR tnt)

ineskhou: also feel free to add commecnts on my part of the tutorial!

pysquared/cdh.py
```diff
@@ -121,17 +128,16 @@ def listen_for_commands(self, timeout: int) -> None:
...
         self._packet_manager.send(joke.encode("utf-8"))
 
     def change_radio_modulation(self, args: list[str]) -> None:
-        """
-        Change the radio modulation.
-        :param modulation: The new radio modulation to set.
+        """Changes the radio modulation.
+
+        Args:
+            args: A list of arguments, where the first argument is the new modulation.
```
ineskhou, line 140: could be nice to have explaination of what other arguments could be/link to docs?

docs/design-guide.md
nateinaction: All todos are gone now! :-)

docs/getting-started.md
nateinaction: I think it's a great idea to tell folks about other potential tools they can use. I want this getting started guide to be a bullet-proof walk through that works every time for every system so I want to keep the tooling pretty bare bones.

I think it would be great to have tips for other applications and tooling in another guide that we can link to at the bottom of Getting Started. What do you think?

pysquared/cdh.py
nateinaction: Yeah this is a little non-obvious. I just wrote this a few weeks ago and I had to think about how this worked even with the function documentation right in front of me.

The way this is written has to do with the way commands are sent to the satellite. Right now they are sent as a packet that includes a `command` field and an `args` field. The `args` field is a list of strings and my thought process here is that we don't know in the parent function what kind of validation to apply to the `args` or which `args` index to send to the function that runs the command. Instead we pass the entire `args` list to the command function and let it decide what to do with that information.

In this case there is only one valid argument to the `change_radio_modulation()`, it's the name of the modulation. All other values passed to `args` would be ignored.


What about something like this?
```suggestion
            args: A list of arguments, the first item must be the new modulation. All other items in the args list are ignored.
```

pysquared/hardware/magnetometer/manager/lis2mdl.py
```diff
@@ -1,30 +1,40 @@
...
         i2c: I2C,
     ) -> None:
-        """Initialize the manager class.
+        """Initializes the LIS2MDLManager.
 
-        :param Logger logger: Logger instance for logging messages.
-        :param busio.I2C i2c: The I2C bus connected to the chip.
+        Args:
+            logger: The logger to use.
+            i2c: The I2C bus connected to the chip.
```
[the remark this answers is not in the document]
nateinaction: This is true. I generated the docs before writing the design guide. I would fix it but I'm out of free gemini credit. All the same we should see if this really makes a difference in the mkdocs output...

**With (type):**
Captura de pantalla 2025-07-14 a la(s) 22 30 47

**Without (type):**
Captura de pantalla 2025-07-14 a la(s) 22 29 41

It looks like the parameters are rendered for both in mkdocs and the types are already in the function signatures. It feels like we don't need the type redeclared in the comments. Removed in [042e648](https://github.com/proveskit/pysquared/pull/271/commits/042e648f88fb82a80926586ebd500ce876dd3401).

docs/getting-started.md
ineskhou: I personally found tabby to be easier for new members bc everyone regardless of OS could do it and it automatically detects the board so you don't need to look for it. That being said understanding under the hood how everything works is super important and yours reviews that better so for a first lookthrough it makes sense as well! 

I think in an ideal world the tutorials are the same so that everyone is on the same page and they aren't confused if they do one and then the other. Thoughts on me making a tabby page? I can link it from these docs to the tabby page, and the hardware tutorial can link here. So people can go hardware --> software --> tabby if they want but its clear what is what

pysquared/cdh.py
ineskhou: perfect!

Mikefly123 (approved):
LGTM!

Mikefly123 (approved):
LGTM! Kachunk

PR #272, opened 2025-07-14
Add identifiable packets through a counter

hrfarmer:
## Summary
Adds a 1 byte number to the header of packets that identifies the message it came from. Receive will bounce any packets that don't match the identifier it's looking for. Should be a good start for adding retry logic in the future.

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)

### DIFF
--- a/pysquared/hardware/radio/packetizer/packet_manager.py
+++ b/pysquared/hardware/radio/packetizer/packet_manager.py
@@ -172,10 +172,10 @@ class PacketManager:
 
         return b"".join(self._get_payload(packet) for packet in sorted_packets)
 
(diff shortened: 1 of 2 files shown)


hrfarmer:
The number is incremented on each unique message sent, so unless we have a situation where there's over 255 things pending to be sent at a time, it *shouldn't* be an issue.

Also yes for things like file transfer the sequence number should be able to handle that.

nateinaction (approved):

nateinaction (changes requested):
Captura de pantalla 2025-07-15 a la(s) 19 01 45
Looks like we need a test for if we get another message while receiving a message.

PR #299, opened 2025-08-06
Sdcard logging

asiemsen:
## Summary
Created sdCardManager class to initialize and mount sd card. The logger takes in the sd card to ensure that it is mounted and writes all logs into the sd card at /sd/logs/activity.log.

This PR creates a new boot.py that mounts the root filesystem as not readonly, making external changes not possible. This can maybe be changed to only mount the /sd directory instead of the root. 

## How was this tested
- No tests

### DIFF
--- a/pysquared/logger.py
+++ b/pysquared/logger.py
@@ -168,17 +168,17 @@ class Logger:
             )
 
         if self._can_print_this_level(level_value):
-            # Write to sd card if mounted
-            if self.sd_card and self.sd_card.mounted:
-                with open("/sd/logs/log", "a") as f:
-                    f.write(json_output + "\n")
-            else:
-                print("sd card not mounted")
-
             if self.colorized:
                 json_output = json_output.replace(
                     f'"level": "{level}"', f'"level": "{LogColors[level]}"'
                 )
+            # Write to sd card if mounted
+            if self.sd_card and self.sd_card.mounted:
+                with open("/sd/logs/log.log", "a") as f:
+                    f.write(json_output + "\n")
+            else:
+                print("sd card not mounted")
+
             print(json_output)
 
     def debug(self, message: str, **kwargs: object) -> None:
--- a/pysquared/logger.py
+++ b/pysquared/logger.py
@@ -80,7 +80,7 @@ class Logger:
     def __init__(
         self,
         error_counter: Counter,
-        sd_card: SDCardManager,
+        sd_card: SDCardManager = None,
         log_level: int = LogLevel.NOTSET,
         colorized: bool = False,
     ) -> None:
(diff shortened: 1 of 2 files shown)


pysquared/sd_card.py
```diff
@@ -0,0 +1,45 @@
...
+            vfs = storage.VfsFat(sd)
+            storage.mount(vfs, mount_path)
+        except (OSError, ValueError) as e:
+            print("error mounting sd card", e)
+        else:
+            self.mounted = True
+
+        if "logs" not in os.listdir("/sd"):
+            print("/sd/logs does not exist, creating...")
+            os.mkdir("/sd/logs")
```
nateinaction, line 45: Since `SD` is our generic manager for handling SD card mounting, the specific usage of the SD card by the logger should likely live in the logger.

pysquared/sd_card.py
```diff
@@ -0,0 +1,45 @@
...
+        self.log_rotate = log_rotate
+
+        try:
+            sd = sdcardio.SDCard(spi_bus, chip_select, baudrate)
+            vfs = storage.VfsFat(sd)
+            storage.mount(vfs, mount_path)
+        except (OSError, ValueError) as e:
+            print("error mounting sd card", e)
+        else:
+            self.mounted = True
```
nateinaction, line 41: Instead of using the concept of `mounted=True` let's raise a hardware initialization error [like other managers](https://github.com/proveskit/pysquared/blob/bc7eb68ed9e80ec7ac3945502eb8e0acafad9e89/pysquared/hardware/imu/manager/lsm6dsox.py#L51C19-L52C81).

pysquared/logger.py
```diff
@@ -79,6 +80,7 @@ class Logger:
     def __init__(
         self,
         error_counter: Counter,
+        sd_card: SDCardManager = None,
```
nateinaction, line 83: Yeah, I was wrong about this. Instead of accepting an SDCard here we should accept [a path](https://docs.circuitpython.org/projects/pathlib/en/latest/). If a path is set, we should verify it exists during the init if it doesn't raise an error.

nateinaction (commented):
This is a great start! Congrats getting this PR up, I know it's been in the making for a while!

I've made some suggestions that I hope will improve separation of concerns between the SD Card Manager and the Logger classes. Let me know what you think!

pysquared/logger.py
asiemsen: Hey Nate, first I just wanted to say thank you for all of your help so far with this PR! I'm finally about done getting it updated, but was having some trouble understanding the intentions of this recommendation. Because the sd card cannot be initialized before the logger, if the path is ever set on initialization (ex: sd_path = sd/logs) it can never exist, or if it does, the sd card will not be mounted to it which I'm not sure what the resulting behavior would be if we tried writing to that path. Maybe there could be a function within the logger to set the sd_path, verifying it exists when set? Another thing that I wasn't sure of was when verifying if it exists, would we then want to create the new directory and then raise an error, sort of similar to what is implemented now?:
```
if "logs" not in os.listdir("/sd"):
            print("/sd/logs does not exist, creating...")
            os.mkdir("/sd/logs")
```
Not sure if I am just misunderstanding how the path works or just the implementation you are imagining with the comment. 

Thanks again for all the help!

nateinaction (approved):

PR #300, opened 2025-08-10
Adding licence to packets so every packet has the senders identify

ineskhou:
## How was this tested
- [ ] Added new unit tests
-  Ran code on hardware (screenshots are helpful) 
- [ ] Other (Please describe)


Mikefly123:
Hey @ineskhou, when you have the time to revisit this, I think adding callsigns to the messages is a good idea although it appears to be completely blowing up all the tests.

PR #302, opened 2025-08-19
Bit Packing

yudataguy:
# Binary Bit Packing Implementation Summary

## Overview

I implemented a binary encoding system that replaces JSON string representations with compact binary encoding for int, float, and string values, achieving 66.6% packet size reduction (784 bytes → 262 bytes).

 ## Core Architecture

1. BinaryEncoder Class (pysquared/binary_encoder.py)

  class BinaryEncoder:
      def add_int(key, value, size=4)     # 1, 2, 4, or 8 byte integers
      def add_float(key, value, precision) # 4 or 8 byte IEEE 754 floats
      def add_string(key, value)          # Length-prefixed UTF-8 strings
      def to_bytes()                      # Generate compact binary format

2. Binary Format Design

  Each field uses this compact structure:
  [key_hash:4][type:1][data:variable]

  - Key Hash (4 bytes): CRC32 hash of field name instead of storing full string
  - Type Identifier (1 byte):
    - 0 = String
    - 1 = 1-byte signed/unsigned int
    - 2 = 2-byte signed/unsigned int
    - 3 = 4-byte signed/unsigned int
    - 4 = 8-byte signed/unsigned int
    - 5 = 4-byte IEEE 754 float
    - 6 = 8-byte IEEE 754 double
  - Data (variable): Raw binary representation

3. Intelligent Size Optimization

```
  ### Automatic integer size selection based on value range
  if -128 <= value <= 127:     encoder.add_int(key, value, size=1)  # 1 byte
  elif -32768 <= value <= 32767: encoder.add_int(key, value, size=2)  # 2 bytes  
  else:                        encoder.add_int(key, value, size=4)  # 4 bytes
```

  ## Integration Points

  1. Beacon Service Update (pysquared/beacon.py)

```
  ### OLD: JSON encoding
  b = json.dumps(state, separators=(",", ":")).encode("utf-8")

  ### NEW: Binary encoding  
  b = self._encode_binary_state(state)

  def _encode_binary_state(self, state):
      encoder = BinaryEncoder()
      for key, value in state.items():
          if isinstance(value, int):
              # Smart size selection
          elif isinstance(value, float):
              encoder.add_float(key, value)
          elif isinstance(value, (list, tuple)):
              # Handle sensor arrays (IMU data)
          else:
              encoder.add_string(key, str(value))
      return encoder.to_bytes()
```

  2. Decoder for Received Data

```
  @staticmethod
  def decode_binary_beacon(data: bytes, key_map: dict = None) -> dict:
      decoder = BinaryDecoder(data, key_map)
      return decoder.get_all()
```

  Technical Innovations

  1. Hash-Based Key Storage

  Instead of storing full field names like "PowerMonitor_0_current_avg" (25 bytes), we store a 4-byte hash:
  `key_hash = hash(key) & 0xFFFFFFFF  # 4-byte hash`

  2. Variable Integer Encoding

  ### Battery level: 85 (fits in 1 byte)
  struct.pack(">IBB", key_hash, 1, 85)  # 6 bytes total

  ### Timestamp: 1705315845 (needs 4 bytes)  
  struct.pack(">IBI", key_hash, 3, 1705315845)  # 9 bytes total

  3. IEEE 754 Float Packing

  ### Temperature: 23.5°C (4-byte float)
  struct.pack(">IBf", key_hash, 5, 23.5)  # 9 bytes total

  ### High precision: (8-byte double)
  struct.pack(">IBd", key_hash, 6, 3.141592653589793)  # 13 bytes total

  4. Array Handling for Sensor Data

```
  ### IMU acceleration [x, y, z] arrays become individual fields:
  ### "IMU_0_acceleration" → "IMU_0_acceleration_0", "IMU_0_acceleration_1", "IMU_0_acceleration_2"
  for i, v in enumerate(acceleration_array):
      encoder.add_float(f"{key}_{i}", float(v))
```

  Performance Results

  | Metric            | JSON      | Binary    | Improvement     |
  |-------------------|-----------|-----------|-----------------|
  | Packet Size       | 784 bytes | 262 bytes | 66.6% reduction |
  | Compression Ratio | 1.0x      | 2.99x     | 3x smaller      |
  | Daily Bandwidth   | 2.2 MB    | 754 KB    | 1.5 MB saved    |
  | Per Transmission  | 790 bytes | 268 bytes | 522 bytes saved |

  Key Benefits

  1. Space Efficiency: 4-byte hashes vs full field names
  2. Type-Aware Encoding: Optimal byte usage per data type
  3. IEEE 754 Compliance: Standard float representation
  4. Backward Compatibility: send_json() method preserved
  5. Robust Decoding: Handles malformed/truncated data gracefully
  6. CircuitPython Compatible: No external dependencies

  Example Encoding Comparison

  JSON (784 bytes):
  `{"name":"PySquared-1","time":"2024-08-19 14:30:45","uptime":7256.3,"Processor_0_temperature":42.8,"battery_level":87,...}`

  Binary (262 bytes):
 ` [9C142B15][00][0B]PySquared-1[4BC95CD2][05][45E31A66][A25F8C14][05][425F999A]...`

The binary format achieves massive space savings by eliminating JSON overhead, using fixed-width type identifiers, and leveraging efficient binary representations for all numeric data.

### DIFF
--- a/cpython-workspaces/flight-software-unit-tests/src/unit-tests/test_beacon.py
+++ b/cpython-workspaces/flight-software-unit-tests/src/unit-tests/test_beacon.py
@@ -1041,13 +1041,3 @@ def test_beacon_encode_sensor_dict_with_non_numeric_values():
     # Verify encoding completed without error
     encoded_data = encoder.to_bytes()
     assert isinstance(encoded_data, bytes)
-
-
-def test_typing_import_exception_handling():
-    """Tests the import exception handling for typing.OrderedDict (lines 36-37)."""
-    # This test verifies that the import exception is handled gracefully
-    # The actual import happens at module load time, so we can't easily test
-    # this without reloading the module. Instead, we verify the code structure
-    # allows for the exception to be caught silently.
-    # The fact that beacon.py loads successfully proves this works
-    assert True
--- a/cpython-workspaces/flight-software-unit-tests/src/unit-tests/test_binary_encoder.py
+++ b/cpython-workspaces/flight-software-unit-tests/src/unit-tests/test_binary_encoder.py
@@ -412,3 +412,67 @@ class TestBeaconIntegration:
         assert savings_percent > 0, (
             f"Should have positive savings, got {savings_percent:.1f}%"
         )
+
+    def test_get_int_size_edge_cases(self):
+        """Test _get_int_size method for different value ranges."""
+        encoder = BinaryEncoder()
+
+        # Test 2-byte range (line 73)
+        encoder.add_int("medium_pos", 32767)  # Max for 2 bytes
+        encoder.add_int("medium_neg", -32768)  # Min for 2 bytes
+
+        # Test 8-byte range (line 77)
+        encoder.add_int("large_pos", 2147483648)  # Requires 8 bytes
+        encoder.add_int("large_neg", -2147483649)  # Requires 8 bytes
+
+        data = encoder.to_bytes()
+        decoder = BinaryDecoder(data, encoder.get_key_map())
+
+        assert decoder.get_int("medium_pos") == 32767
+        assert decoder.get_int("medium_neg") == -32768
+        assert decoder.get_int("large_pos") == 2147483648
+        assert decoder.get_int("large_neg") == -2147483649
+
+    def test_decoder_truncated_data_edge_cases(self):
+        """Test decoder with various truncated data scenarios."""
+        encoder = BinaryEncoder()
+        encoder.add_string("test", "hello")
+        encoder.add_int("num", 42)
+        full_data = encoder.to_bytes()
+
+        # Test truncated string length byte (line 322)
+        # Need exactly key_hash (4 bytes) + type (1 byte) but no length byte
+        truncated_at_string_len = full_data[
+            :5
+        ]  # Just key_hash + type, no string length
+        decoder = BinaryDecoder(truncated_at_string_len, encoder.get_key_map())
+        result = decoder.get_all()
+        # Should handle the truncation gracefully
+        assert isinstance(result, dict)  # May be empty or partial
+
+        # Test truncated numeric data (line 348)
+        encoder2 = BinaryEncoder()
+        encoder2.add_int("test_int", 123456)  # 4-byte int
+        int_data = encoder2.to_bytes()
+        # Truncate so we have key_hash + type but not enough bytes for the int
+        truncated_numeric = int_data[
+            :7
+        ]  # key_hash(4) + type(1) + partial_data(2 of 4 needed)
+        decoder2 = BinaryDecoder(truncated_numeric, encoder2.get_key_map())
+        result2 = decoder2.get_all()
+        # Should handle truncation gracefully
+        assert isinstance(result2, dict)  # May be empty
+
+    def test_corrupted_string_data(self):
+        """Test decoder with corrupted string data."""
+        # Create data that claims to have a string but is truncated
+        key_hash = b"\x12\x34\x56\x78"  # 4-byte key hash
+        string_type = b"\x00"  # String type = 0
+        string_length = b"\x10"  # Claims string is 16 bytes long
+        partial_string = b"hello"  # But only 5 bytes provided
+
+        corrupted_data = key_hash + string_type + string_length + partial_string
+        decoder = BinaryDecoder(corrupted_data)
+        result = decoder.get_all()
+        # Should handle corruption gracefully without crashing
+        assert isinstance(result, dict)


pysquared/beacon.py
```diff
@@ -86,15 +88,66 @@ def send(self) -> bool:
...
+            Binary encoded data
+        """
+        encoder = BinaryEncoder()
+
+        for key, value in state.items():
+            self._encode_value(encoder, key, value)
+
+        return encoder.to_bytes()
+
+    def _encode_value(self, encoder: BinaryEncoder, key: str, value: object) -> None:
```
Mikefly123, line 112: I think this helper function is a little bulky to include in `beacon.py` and it is increasing the complexity of file by quite a bit. If we want to do the auto-magic detection and encoding let's consider moving this helper into the `binary_encoder.py` so the `beacon` functionality is easier to understand and this functionality can also be used by other services.

I think there is also an argument that maybe we shouldn't do the auto detection for encoding at all and should just expect the user to know what type they are trying to encode when the data is generated and call that encoder module directly. That may actually be a safer approach, since we have control over the entire codebase, rather than trying to do it dynamically.

pysquared/binary_encoder.py
```diff
@@ -0,0 +1,300 @@
...
+        for key, (fmt, value) in self._data.items():
+            # Use a simple hash of the key instead of storing the full key
+            key_hash = hash(key) & 0xFFFFFFFF  # 4-byte hash
+            self._key_map[key_hash] = key
+
+            result += self._encode_field(key_hash, fmt, value)
+
+        return result
+
+    def _encode_field(
```
Mikefly123, line 116: This is a pretty obfuscated function, I think we may want to consider taking each of the nested functionalities and splitting them into helper functions. This would blow up the number of functions by quite a bit, but we could also really improve readability by having this function just be the switch case and the helpers being named for what each block does!

tests/unit/test_binary_encoder.py
```diff
@@ -0,0 +1,391 @@
...
+        assert abs(large_result - 1e10) < 1e5
+
+    def test_unknown_format_error(self):
+        """Test error handling for unknown format in _encode_field."""
+        encoder = BinaryEncoder()
+        # This should not happen in normal usage, but test the error path
+        with pytest.raises(ValueError, match="Unknown format"):
+            encoder._encode_field(123, "X", 42)
+
+    def test_import_fallback(self):
```
Mikefly123, line 286: I'm not sure that this assert is actually needed since there is no logic inside this test block.

Mikefly123 (changes requested):
Hey @yudataguy we are getting close! I'm going to have @ineskhou run a test of this on hardware soon and we should be ready to rock and roll.

I just went ahead and made some suggestions based on the [SonarQube](https://sonarcloud.io/summary/new_code?id=proveskit_circuitpy_flight_software&pullRequest=302) analysis that you should consider making tweaks based on to clean up some of the more confusing parts of the code!

pysquared/beacon.py
```diff
@@ -86,15 +88,140 @@ def send(self) -> bool:
...
+                encoder.add_string(key, str(value))
+        # Handle beacon system data with known types
+        elif key in ("name", "time"):
+            encoder.add_string(key, str(value))
+        elif key == "uptime":
+            encoder.add_float(key, self._safe_float_convert(value))
+        elif "temperature" in key or "voltage" in key or "current" in key:
+            encoder.add_float(key, self._safe_float_convert(value))
+        elif "timestamp" in key:
+            encoder.add_float(key, self._safe_float_convert(value))
```
Mikefly123, line 147: Rather than checking the key for what kind of type the value is, why don't we try just directly checking the type of the value using `isinstance`?
```suggestion
       elif isinstance(value, float)
```

pysquared/beacon.py
```diff
@@ -86,15 +88,140 @@ def send(self) -> bool:
...
+            encoder.add_string(key, str(value))
+        elif isinstance(value, bool):
+            encoder.add_int(key, int(value), size=1)
+        elif isinstance(value, int):
+            # Use appropriate size based on expected range
+            if -128 <= value <= 255:  # Counter values, small readings
+                encoder.add_int(key, value, size=1)
+            elif -32768 <= value <= 32767:  # Medium range values
+                encoder.add_int(key, value, size=2)
+            else:  # Timestamps, large values
```
Mikefly123, line 162: If we are concerned about the complexity of this nested section, we could probably move the size checking into the `add_int` function.

pysquared/beacon.py
```diff
@@ -86,15 +88,140 @@ def send(self) -> bool:
...
+                    # Handle 3D vectors (acceleration, gyroscope)
+                    for i, v in enumerate(dict_value):
+                        encoder.add_float(
+                            f"{full_key}_{i}", self._safe_float_convert(v)
+                        )
+                else:
+                    encoder.add_float(full_key, self._safe_float_convert(dict_value))
+            else:
+                # Handle other sensor-specific fields
+                if isinstance(dict_value, (int, float)):
```
Mikefly123, line 217: I think these can be un-nested from behind the `else` statement and just put on the `elif` chain.

pysquared/beacon.py
```diff
@@ -86,15 +88,140 @@ def send(self) -> bool:
...
+            encoder: The binary encoder to add data to
+            key: The base key name
+            sensor_data: Dictionary containing sensor readings
+        """
+        for dict_key, dict_value in sensor_data.items():
+            full_key = f"{key}_{dict_key}"
+            if dict_key == "timestamp":
+                encoder.add_float(full_key, self._safe_float_convert(dict_value))
+            elif dict_key == "value":
+                if isinstance(dict_value, (list, tuple)) and len(dict_value) == 3:
```
Mikefly123, line 207: This could also potentially be un-nested. Where we can try to directly check if it is `isinstance(dict_value, (list, tuple)) and len(dict_value) == 3:` without checking for the `dict_key == "value":` first. The else block would also move out into an elif check for floats

Mikefly123 (commented):
Hey @yudataguy added some thoughts on how we can tackle those last two complex functions!

ineskhou:
Looks good pending the small logic error, let me know once you've done the readability changes and ill do a robust lookover on the hardware again

nateinaction (commented):
Hey @yudataguy nice work!

It looks like there are is some low hanging fruit to improve our test coverage. Could you take a look? https://sonarcloud.io/component_measures?id=proveskit_circuitpy_flight_software&pullRequest=302&metric=new_coverage&view=list&selected=proveskit_circuitpy_flight_software%3Acircuitpython-workspaces%2Fflight-software%2Fsrc%2Fpysquared

nateinaction:
Updated with new repo structure in [877c37a](https://github.com/proveskit/pysquared/pull/302/commits/877c37a4995af915df75dce2f56658ccb7507a0e). You'll need to `make clean test typecheck` to get set up again. Here is the new contributing guide section on using the new workspaces feature: https://proveskit.github.io/pysquared/contributing/#workspaces

Mikefly123:
Hey @yudataguy! The threshold is still 80% on the codebase, but I think Nate is saying we might be able to hit 100% with just checking for a few more conditions in Beacon and Binary Encoder. I tried to toss an LLM at it right now, but it stumbled a bit on the new workspace structure haha.

I don't think it is very critical to push it all the way up to 100%, but if you have a few minutes to get it done it would be nice. We could also open it as a `good first issue` ticket for any newcomers to the codebase to take a crack at down the line and just merge it in as is. Let me know what you think!

PR #305, opened 2025-09-01
Feedback on SD Card Logging

nateinaction:
## Summary
This PR provides feedback on #299 

## How was this tested
- [x] Added new unit tests
- [ ] Other (Please describe)

`boot_out.txt` first run:
```shell
Adafruit CircuitPython 9.2.8 on 2025-05-28; PROVES Kit v4 with rp2040
Board ID:proveskit_rp2040_v4
UID:E4639C3563112E2D
boot.py output:
Disabled USB drive
Remounted root filesystem
Mount point /poke created.
Enabled USB drive
```

`boot_out.txt` subsequent runs:
```shell
Adafruit CircuitPython 9.2.8 on 2025-05-28; PROVES Kit v4 with rp2040
Board ID:proveskit_rp2040_v4
UID:E4639C3563112E2D
boot.py output:
Disabled USB drive
Remounted root filesystem
Mount point /poke already exists.
Enabled USB drive
```

Demonstrating logging to file:
Captura de pantalla 2025-09-01 a la(s) 19 03 31

### DIFF
--- a/pysquared/logger.py
+++ b/pysquared/logger.py
@@ -77,11 +77,12 @@ class LogLevel:
 class Logger:
     """Handles logging messages with different severity levels."""
 
+    _log_dir: str | None = None
+
     def __init__(
         self,
         error_counter: Counter,
         log_level: int = LogLevel.NOTSET,
-        log_dir: str | None = None,
         colorized: bool = False,
     ) -> None:
         """
@@ -90,25 +91,11 @@ class Logger:
         Args:
             error_counter: Counter for error occurrences.
             log_level: Initial log level.
-            log_dir: Directory to save log files. If None, logs are not saved to a file.
             colorized: Whether to colorize output.
         """
         self._error_counter: Counter = error_counter
         self._log_level: int = log_level
-        self._log_dir: str | None = log_dir
         self._colorized: bool = colorized
-
-        if log_dir is not None:
-            try:
-                # Octal number 0o040000 is the stat mode indicating the file being stat'd is a directory
-                directory_mode: int = 0o040000
-                st_mode = os.stat(log_dir)[0]
-                if st_mode != directory_mode:
-                    raise ValueError(
-                        f"Logging path must be a directory, received {st_mode}."
-                    )
-            except OSError as e:
-                raise ValueError("Invalid logging path.") from e
 
     def _can_print_this_level(self, level_value: int) -> bool:
         """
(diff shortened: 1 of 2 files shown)


pysquared/logger.py
```diff
@@ -18,8 +18,6 @@
 import traceback
 from collections import OrderedDict
 
-import adafruit_pathlib
```
nateinaction, line 21: While I previously suggested using `adafruit_pathlib`, this would have caused additional pain in a few ways:
- All CircuitPython functions that interact with the filesystem expect a string path which would mean that we would always need to output the path to a string anyways.
- `adafruit_pathlib` did not have any `join()` semantics which made adding to a `Path` object clunky. (this could be submitted upstream)
- Most modules in pysquared that use the `Logger`, I found that I would have had to mock away the use of `adafruit_pathlib` in many many test files.

So ultimately, I removed it.

pysquared/beacon.py
```diff
@@ -19,9 +19,9 @@
 from collections import OrderedDict
 
 try:
-    from mocks.circuitpython.microcontroller import Processor
-except ImportError:
     from microcontroller import Processor
+except ImportError:
+    from mocks.circuitpython.microcontroller import Processor
```
nateinaction, line 24: Not related to this PR: switching this order is possible now because of #304. With this order change, the IDE shows more accurate hints.

pysquared/logger.py
```diff
@@ -162,30 +167,15 @@ def _log(self, level: str, level_value: int, message: str, **kwargs) -> None:
...
-        except TypeError as e:
-            json_output = json.dumps(
-                OrderedDict(
-                    [
-                        ("time", asctime),
-                        ("level", "ERROR"),
-                        ("msg", f"Failed to serialize log message: {e}"),
-                    ]
-                ),
-            )
```
nateinaction, line 176: This became no longer necessary after #228 but was left in. It has no tests and is no longer required so I removed it.

tests/unit/hardware/test_burnwire.py
nateinaction, line 1: Moved into a new `burnwire/` dir to match how it's organized in `pysquared/`.

asiemsen (approved):
Looks great! I would love to go over and review the design behind the tests sometime so that I can start implementing some of these in the future! I also saw your most recent commit that uses a setter for the log directory, which I think was implemented really well. Thanks again for all your help on this.

Issue #310, opened 2025-09-28
[BUG] Beacon is Not Detecting Magnetometer Proto


Mikefly123:
## Summary
We're getting a fresh `typecheck` error when downstream implementation repos are updating to use 25w36+ releases. I think this is just a quick patch that is needed to the `Beacon` class.
```sh
  /home/runner/work/CircuitPython_RP2350_v5b/CircuitPython_RP2350_v5b/src/flight-software/main.py:124:9 - error: Argument of type "LIS2MDLManager" cannot be assigned to parameter "args" of type "PowerMonitorProto | RadioProto | IMUProto | TemperatureSensorProto | Flag | Counter | Processor" in function "__init__"
    Type "LIS2MDLManager" is not assignable to type "PowerMonitorProto | RadioProto | IMUProto | TemperatureSensorProto | Flag | Counter | Processor"
      "LIS2MDLManager" is not assignable to "PowerMonitorProto"
      "LIS2MDLManager" is not assignable to "RadioProto"
      "LIS2MDLManager" is not assignable to "IMUProto"
      "LIS2MDLManager" is not assignable to "TemperatureSensorProto"
      "LIS2MDLManager" is not assignable to "Flag"
      "LIS2MDLManager" is not assignable to "Counter"
      "LIS2MDLManager" is not assignable to "Processor" (reportArgumentType)
```
I think the patch for this will be to update `Beacon` with the Magnetometer Proto in the `__init__` and also add a function that parses the magnetometer data in a similar way to the IMU data.
## Steps to reproduce
- Update a downstream implementation repo to use PySquared 25w36+
- Run the typechecker

## Environment
- GitHub Actions CI

Mikefly123:
Hey @yudataguy or @ineskhou if either of you have time to push a quick patch for this bug that would be greatly appreciated! I might have some time later in the week if we don't get to it by Tuesday's meeting.

PR #312, opened 2025-09-29
Fix bug of Beacon with Magnetometer Proto

yudataguy:
## Summary
Fix bug of Beacon with Magnetometer Proto

#310 I'm not sure how to get the downstream updated. Hope there's way to test it before fully PR this


Mikefly123:
Hey @yudataguy! This does seem to fix it! 

The way to test would be to go to one of the implementation repos (like the V4 one) and change the first line of the `Makefile` to point to your branch rather than a release tag. When using `make clean` then `make` it will delete your old libraries and download the new ones (from your branch!). Then a quick `make typecheck` will figure out if the type error still exists:
Screenshot 2025-09-29 at 12 27 27 AM

It seems like I am type checking with no errors now! 
Screenshot 2025-09-29 at 12 30 10 AM

If you can just (maybe vibe code?) some updates to the unit tests to appease SonarCloud we can get this PR merged in.

PR #316, opened 2025-10-01
Add Amateur Radio Repeat Code

Mikefly123:
## Summary
One of the core functionalities we want to the satellites to be able to do is receive a message from radio amateurs and then repeat it! This PR scaffolds that functionality by adding an OSCAR (Orbiting Satellite Carrying Amateur Radio) function to `cdh`. 

More deliberation is needed on how to do this in a secure way.

## How was this tested
- [ ] Added new unit tests
- [ ] Other (Please describe)


Mikefly123:
This PR is now ready for testing! The code should be symmetrical across both the `flight-software` and `ground-station` so all that needs to be done is a quick checkout and test against downstream implementation software.

